PR #2, opened 2016-05-12
Add build-info/dotnet/corefx-tfs/master package version files

dagood:
Adds files to track the latest available packages published out of the TFS build. There should be no dependencies between these ~~`corefx-tfs`~~ `projectk-tfs` packages and the ones in `build-info/dotnet/corefx/master` when we start updating this repo.

/cc @weshaggard @eerhardt @ericstj @jhendrixMSFT


eerhardt:
@weshaggard - I'm not certain what the permissions are for https://github.com/dotnet/corefx, but I think this repo should have the same permissions as dotnet/corefx. Can you either tell me what I should add to the versions repo, or can you add the permissions to the versions repo?

eerhardt:
@dagood - can you write a list of which packages are going to go into this corefx-tfs folder?

dagood:
@jhendrixMSFT might be able to provide a more current list based on ongoing work, but this is what I get by diffing the dotnet/versions master packages vs. dagood/versions master: https://gist.github.com/dagood/3a1271c4ae6da31c59bbc90a8dc5b7f6.

eerhardt:
Do we still need every package in that list?  What are:

```
Microsoft.NETCore.Console
Microsoft.NETCore.ConsoleHost
Microsoft.NETCore.TestHost
```

those packages for?  Can't they be replaced with the Microsoft.DotNetHost (i.e. the `dotnet` native executable) packages?

dagood:
Chatted with @eerhardt about this, and now I think `projectk-tfs` is a clearer name for this new group, since the packages that would be in `corefx-tfs` don't seem related to `corefx`. (Other ideas: `core-tfs`, `dotnet-tfs`, `projectk`.) Renamed it in this PR, but open for thoughts.

eerhardt:
I like `projectk-tfs` - hopefully it should be going away over time as `coreclr` and `wcf` start building from git as wel.

eerhardt:
I'm merging this.  We can continue the conversation and make changes as we go.

@weshaggard - can you help with the permissions to the dotnet/versions repo?

weshaggard:
I'm ok with the projectk-tfs name as well, especially given that I hope it is short lived. 

@eerhardt as for permissions I wouldn't add more permissions to this repo. Most people shouldn't need to commit to this repo as dotnet-bot should be doing most of that work. The trouble I have with making it match corefx is that it won't be enough because we expect this to be used by all of dotnet eventually so we would need to give everyone access to it. For now lets just keep it the way it is.

eerhardt:
OK - sounds good.  Thanks, Wes.

PR #11, opened 2016-05-31
Add CoreFX subscriptions for ProjectK TFS build

dagood:
Adds a subscription that updates the new dependency version categories that will be created in https://github.com/dotnet/corefx/pull/8999.

~~It's a bit repetitive. `UpdateDependencies.ps1` could be updated to allow for an array of elements to update, but doing it this way makes it easier to split them off later.~~ (Added array support.)

/cc @eerhardt @weshaggard

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -59,7 +59,7 @@
     {
       "path": "https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/Latest.txt",
       "handlers": [
-        // This handler will bring the Latest CoreFX master build into CorefX master
+        // This handler will bring the Latest CoreFX master build into CoreFX master
         {
           "maestroAction": "corefx-general",
           "maestroDelay": "00:10:00",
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -69,7 +69,55 @@
...
+        // This handler will bring the Latest ProjectK TFS master build into CoreFX master for CoreCLR dependencies
+        {
+          "maestroAction": "corefx-general",
+          "maestroDelay": "00:10:00",
+          "ScriptFileName": "UpdateDependencies.ps1",
+          "Arguments": [
+            "-GitHubUser dotnet-bot",
+            "-GitHubEmail dotnet-bot@microsoft.com",
+            "-GitHubPassword `$(`$Secrets['DotNetBotGitHubPassword'])",
+            "-CoreFxVersionUrl https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/projectk-tfs/master/Latest.txt",
```
weshaggard: This looks like it is going to cause 3 PR's to corefx whenever we update projectk-tfs. I suspect you are just setting them up independently since we are going they will be independent soon.

Maestro/subscriptions.json
```diff
@@ -69,7 +69,55 @@
...
+            "-GitHubPullRequestNotifications 'dotnet/corefx-contrib",
```
weshaggard: I think ExternalExpectedPrerelease and CoreLibTargetingPackExpectedPrerelease can be merged into just External for now until we have a plan to split them which I don't think we do right now.

Maestro/subscriptions.json
dagood: You're right, and on second thought I think adding `DirPropsVersionElement` array functionality to the script is justified to reduce this.

PR #12, opened 2016-06-01
Added subscription to trigger building dotnet-preview Docker image

MichaelSimons:
This happens whenever a new CLI build is completed.  The referenced build definition simply makes the appropriate http put request to Docker Hub that will trigger the build.  Maybe at some time in the future Maestro will have support for this type of action rather than needing to use a VSO build definition.

@eerhardt - not sure how to test this change.  Please advise.

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -24,6 +24,12 @@
       "vsoProject": "DevDiv",
       "buildDefinitionId": 808
     }
+    // A build definition that will trigger the dotnet-preview Docker image build in Docker Hub
+    "build-dotnet-docker-preview": {
+      "vsoInstance": "mseng.visualstudio.com",
+      "vsoProject": "dotnetcore",
+      "buildDefinitionId": 3590
+    }
   },
   "subscriptions": [
     {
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -83,6 +89,16 @@
           "maestroDelay": "00:05:00"
         }
       ]
+    },
+    {
+      "path": "https://github.com/dotnet/versions/blob/master/build-info/dotnet/cli/rel/1.0.0/Latest_Packages.txt",
```
eerhardt: Latest_Packages.txt gets updated by the Windows x64 build right now, since that is what pushes the packages to MyGet.

Is this what you want to respond to?

eerhardt: You probably need @sokket's change and respond to the "build.version" file:  https://github.com/dotnet/cli/pull/3270/files#diff-09865b5728dc444ab9bdd51db40314efR112

MichaelSimons: Agreed, this needs the change you mentioned.  I just wanted to get the ball rolling on this :smile: The current way the docker builds are getting triggered are lagging behind the tip builds as well because they are getting triggered off the Ubuntu builds rather than in FinalizeBuild.  I will update when https://github.com/dotnet/cli/pull/3270 is completed.

eerhardt:
:shipit:

Issue #24, opened 2016-06-09
Fix CoreFXGeneralExecutor-Release failures due to CoreFX native build failing


dagood:
Some dependency updates fail because the CoreFX native build doesn't work:

* https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=206187
* https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=205575
* https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=205170

These dependency updates probably shouldn't be doing a native build at all.

dagood:
Two options for a fix:

1. Add `managed` arg to `build.cmd` call: https://github.com/dotnet/corefx/blob/release/1.0.0/UpdateDependencies.ps1#L56.
2. Set `__buildspec` to `managed` before calling `UpdateDependencies.ps1`.

This doesn't seem to be an issue in CoreFX master, but we should apply a fix there too because the native build is pointless and has the risk of failing the auto-update unnecessarily.

/cc @jhendrixMSFT @eerhardt

PR #25, opened 2016-06-10
Fix VersionFileUrl parameters to get raw github content

dagood:
With the non-raw GitHub url, the update script was adding random HTML into `dir.props`, which fails the auto-PR build. This only happened for recent release builds because when I made those I copy-pasted the subscription `path` into the `VersionFileUrl` parameter without spotting that critical difference.

Fixes https://github.com/dotnet/versions/issues/24, although we should still stop doing native builds in the auto-update build just because it's unnecessary.

/cc @jhendrixMSFT @eerhardt @joshfree


eerhardt:
LGTM.  Want me to merge?

dagood:
Yep, that would be great. :smile:

PR #35, opened 2016-06-17
Remove ConsoleHost package to match feed state

dagood:
There is now a stub package available so that restores don't fail while these dependencies are removed, but this package isn't being shipped.

/cc @weshaggard @eerhardt


eerhardt:
:shipit:

weshaggard:
LGTM - Thanks.

PR #39, opened 2016-07-19
Add build-info and subscriptions for corefx dev/api branch

dagood:
Adds blank build-info files to the versions repo. Replicated the CoreFX master subscriptions and adjusted for this branch.

/cc @eerhardt @danmosemsft @ellismg @stephentoub

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -1,6 +1,6 @@
 {
   "actions": {
-    // A build definition capable of running any .ps1 script in the CoreFX repo master branch
+    // A build definition capable of running any .ps1 script in the CoreFX repo. By default, the master branch.
     "corefx-general": {
       "vsoInstance": "devdiv.visualstudio.com",
       "vsoProject": "DevDiv",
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -6,6 +6,13 @@
       "vsoProject": "DevDiv",
       "buildDefinitionId": 1084
     },
+    // A build definition capable of running any .ps1 script in the CoreFX repo dev/api branch
+    "corefx-general-dev-api": {
+      "vsoInstance": "devdiv.visualstudio.com",
+      "vsoProject": "DevDiv",
+      "vsoSourceBranch": "dev/api",
```
dagood: @eerhardt Did I set this properly? Great to be able to do this instead of cloning the def again. :+1:

eerhardt: No, sorry on the lack of doc here.

The `vsoSourceBranch` goes into the under the "subscriptions" section.  You don't need to create a new "action" here.

See https://github.com/dotnet/versions/blob/master/Maestro/test/test-subscriptions.json for an example.

eerhardt: We could change Maestro's code to look in both places, if you think this way is more understandable.

Just changing it here:  https://github.com/dotnet/versions/blob/master/Maestro/src/Microsoft.DotNet.Maestro/Handlers/VsoBuildHandler.cs#L47-L53 to look in both places.

danmosemsft:
LGTM (doesn't mean much in this case)

Maestro/subscriptions.json
dagood: Ah, I actually think putting it in the subscription is better. Adding a new action was just my first thought because I was expecting to have to clone the definition, and when I saw the `vsoSourceBranch` commit I changed it to this. Fixing.

dagood:
@eerhardt Moved the `vsoSourceBranch` to the subscription, does it look ok now? (On my side, this is ready to merge.)

PR #40, opened 2016-07-19
Add subscriptions to automatically update coreCLR dependencies

wtgodbe:
CC @dagood @eerhardt

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -82,6 +82,20 @@
             "-GitHubPullRequestNotifications 'dotnet/corefx-contrib'",
             "-DirPropsVersionElements CoreFxExpectedPrerelease"
           ]
+        },
+        // This handler will bring the Latest CoreFX master build into CoreCLR master
+        {
+          "maestroAction": "coreclr-general",
+          "maestroDelay": "00:10:00",
+          "ScriptFileName": "UpdateDependencies.ps1",
+          "Arguments": [
+            "-GitHubUser dotnet-bot",
+            "-GitHubEmail dotnet-bot@microsoft.com",
+            "-GitHubPassword `$(`$Secrets['DotNetBotGitHubPassword'])",
+            "-VersionFileUrl https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/corefx/master/Latest.txt",
+            "-GitHubPullRequestNotifications 'dotnet/coreclr-contrib'",
+            "-DirPropsVersionElements CoreFxExpectedPrerelease"
+          ]
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -82,6 +82,20 @@
...
+          "maestroAction": "coreclr-general",
+          "maestroDelay": "00:10:00",
+          "ScriptFileName": "UpdateDependencies.ps1",
+          "Arguments": [
+            "-GitHubUser dotnet-bot",
+            "-GitHubEmail dotnet-bot@microsoft.com",
+            "-GitHubPassword `$(`$Secrets['DotNetBotGitHubPassword'])",
+            "-VersionFileUrl https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/corefx/master/Latest.txt",
+            "-GitHubPullRequestNotifications 'dotnet/coreclr-contrib'",
+            "-DirPropsVersionElements CoreFxExpectedPrerelease"
```
eerhardt: I don't see either `CoreFxExpectedPrerelease` or `CoreClrExpectedPrerelease` in [https://github.com/dotnet/coreclr/blob/master/dir.props](https://github.com/dotnet/coreclr/blob/master/dir.props).  I only see `<ExternalExpectedPrerelease>beta-24319-00</ExternalExpectedPrerelease>`, which already has a handler in this file.

Are we sure this is the right handler?

Maestro/subscriptions.json
```diff
@@ -184,6 +198,20 @@
             "-GitHubPullRequestNotifications 'dotnet/corefx-contrib'",
             "-DirPropsVersionElements CoreClrExpectedPrerelease"
           ]
+        },
+        // This handler will bring the Latest CoreCLR master build into CoreCLR master
```
eerhardt: What in CoreCLR references an earlier version of CoreCLR?  I know why CoreFX needs to be updated with newer CoreFX builds, but what does CoreCLR have a self-dependency?

eerhardt:
:clock1:

eerhardt:
@dagood is added to the review. #Closed

dagood:
Some context for this PR: this depends on https://github.com/dotnet/coreclr/pull/6323, which checks both `dir.props` and `tests/dir.props` for elements to upgrade and adds `CoreFxExpectedPrerelease` and `CoreClrExpectedPrerelease` in `tests/dir.props`.

Maestro/subscriptions.json
dagood: This lets CoreCLR build tests based on an existing CoreCLR package, rather than only the just-built CoreCLR. It's purely a test dependency. (The CoreFX auto-upgrade is also purely for tests.)

Might be useful to clarify in an extended build-info--whether a certain dependency is a product dependency or a test/utility dependency.

---
In reply to: [71458883](https://github.com/dotnet/versions/pull/40#discussion_r71458883) [](ancestors = 71458883)

eerhardt:
Ok - makes sense.  Let me know when I should merge this.

wtgodbe:
Yeah, sorry, I should have been more clear in my description that this depends on https://github.com/dotnet/coreclr/pull/6323. Once that's finished & merged we should be good to go on this PR.

wtgodbe:
@eerhardt can you merge this?

PR #41, opened 2016-07-20
Make upgrade PRs for dev/api against dev/api, not the default master.

dagood:
https://github.com/dotnet/versions/pull/39 used the default value from https://github.com/dotnet/corefx/blob/master/UpdateDependencies.ps1, which meant the PR was for dev/api commits with the upgrade commit on top.

Example PR: https://github.com/dotnet/corefx/pull/10161

/cc @eerhardt @stephentoub


ellismg:
LGTM

PR #43, opened 2016-07-25
Temporarily remove coreclr auto-updates

dagood:
/cc @eerhardt 

(To be reverted when https://github.com/dotnet/buildtools/issues/835 is done and flowed through.)


gkhanna79:
@dagood I take it that this does not disable auto-updates for closed build artifacts. Correct?

dagood:
@gkhanna79 This did disable that too. I've submitted a new PR to add that one back in. (I suppose it is reasonable to keep since it's less frequent than the other builds.)

PR #47, opened 2016-08-05
Revert two test changes that got sent to dotnet/versions, not dagood/versions

dagood:
@eerhardt I was bringing up the new versiontools in coreclr and testing out a command when my config didn't take and the defaults took over, sending the change here rather than to my own repo. This reverts those changes.

I'm not happy with the "default to live" settings in all these versions repo tools... dangerous. (For anyone who has push permissions, like the unfortunately-timed me.) I'm going to rip out the defaults entirely from buildtools unless you think there was a good reason to have them that way.


eerhardt:
:shipit:

PR #50, opened 2016-08-09
Escape quotes after /p:NotifyGitHubUsers= to avoid powershell arg behavior

dagood:
Without this it ends up calling:

    [VSTS inline script].ps1 ... -Arguments "-- ... /p:NotifyGitHubUsers="dotnet/corefx-contrib""

This ignores the "extra" stuff after what it sees as the closing `"` after `=`. I'm not sure what powershell is doing with it that makes this not a syntax error. Escaping the `"` makes it not close the string.

Does this look right, @eerhardt?

I want to keep the quotes because this can become a list by adding `;` and I don't want someone else to potentially have to figure it out later.

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -93,7 +93,7 @@
             "/p:ProjectRepoOwner=dotnet",
             "/p:ProjectRepoName=corefx",
             "/p:ProjectRepoBranch=master",
-            "/p:NotifyGitHubUsers=\"dotnet/corefx-contrib\""
+            "/p:NotifyGitHubUsers=`\"dotnet/corefx-contrib`\""
           ]
         }
       ]
@@ -136,7 +136,7 @@
             "/p:ProjectRepoOwner=dotnet",
             "/p:ProjectRepoName=corefx",
             "/p:ProjectRepoBranch=master",
-            "/p:NotifyGitHubUsers=\"dotnet/corefx-contrib\""
+            "/p:NotifyGitHubUsers=`\"dotnet/corefx-contrib`\""
           ]
         },
         // This handler will bring the Latest ProjectK TFS master build into CoreFX dev/api
@@ -225,7 +225,7 @@
             "/p:ProjectRepoOwner=dotnet",
             "/p:ProjectRepoName=corefx",
             "/p:ProjectRepoBranch=master",
-            "/p:NotifyGitHubUsers=\"dotnet/corefx-contrib\""
+            "/p:NotifyGitHubUsers=`\"dotnet/corefx-contrib`\""
           ]
         },
         // This handler will bring the Latest CoreCLR master build into CoreFX dev/api


eerhardt:
I *think* this looks right.  :shipit:

jaredpar:
I think the JSON is properly escaped as is.  Seems like the bug is else where ...

eerhardt:
> I think the JSON is properly escaped as is. Seems like the bug is else where ...

A thing to know is that this command gets executed by a powershell script in VSO.  See the powershell build step in https://devdiv.visualstudio.com/DevDiv/_build/explorer?_a=simple-process&definitionId=1084

jaredpar:
So many levels of quotes there, hard to see the problem 😦

eerhardt:
It is this line:

```
# evaluate any of the PS expressions in the Arguments (to resolve the secret values)
$Arguments = Invoke-Expression "`"$Arguments`""
```

The Arguments are being evaluated before executing the command.  That is how these commands get access to secret variables in the VSO build.  

The use case is where we have lines like this:
https://github.com/dotnet/versions/blob/master/Maestro/subscriptions.json#L92

```
            "/p:GitHubAuthToken=`$(`$Secrets['DotNetBotGitHubPassword'])",
```

$Secrets is a dictionary filled with all the secret VSO variables.  The Invoke-Expression call turns this line into:

```
            "/p:GitHubAuthToken=12345",
```

where `12345` is the actual GitHub access token to use.

dagood:
I've kicked off a corefx build to verify this, in any case.

That's good info about the `Invoke-Expression`, but just want to make sure it's clear that in this specific case the problem happens when calling the powershell script. I just realized it's actually treating the "ignored" part like a second parameter:

	PS C:\Users\dagood> function foo($a) { Write-Host $a }
	PS C:\Users\dagood> foo "test"str""
	test
	PS C:\Users\dagood> function bar($a, $b) { Write-Host $a; Write-Host $b }
	PS C:\Users\dagood> bar "test"str""
	test
	str

vs.

	PS C:\Users\dagood> foo "test`"str`""
	test"str"

(The json escaping is just to deliver the string to the build.)

dagood:
Ha! It turns out the `Invoke-Expression` is another layer of problem once I get through the original one:

    PS> $command = "test`"inner`""
    PS> Invoke-Expression "`"$command`""
    Invoke-Expression : At line:1 char:7
    + "test"inner""
    +       ~~~~~~~
    Unexpected token 'inner""' in expression or statement.

    PS> $command = "test```"inner```""
    PS> Invoke-Expression "`"$command`""
    test"inner"

So these lines would have to be:

    "/p:NotifyGitHubUsers=```\"dotnet/corefx-contrib```\""

That seems like it goes too far to me, though, so I'm submitting a PR for:

    "/p:NotifyGitHubUsers=dotnet/corefx-contrib"

PR #51, opened 2016-08-09
Disable projectk-tfs release/1.0.0 (KRel) -> CoreFX release/1.0.0 auto-upgrade

dagood:
@gkhanna79 This disables PRs like https://github.com/dotnet/corefx/pull/10663, which are unnecessary right now and fail.

If we start servicing any External packages that need to be referenced in CoreFX, this PR can be reverted.


gkhanna79:
Lgtm

PR #52, opened 2016-08-10
Remove quotes from NotifyGitHubUsers to avoid excessive escaping

dagood:
There's no need to notify multiple users, so this removes my earlier attempt to escape `"`'s to make it possible. If we want to notify multiple users later, it can be done in the `dir.props` of the respective repo along with the upgrade config. It could be done with a bunch of escaping but it's overly confusing.

See https://github.com/dotnet/versions/pull/50 for details about the escaping problem. (cc @eerhardt @jaredpar)


eerhardt:
Looks good to me

PR #53, opened 2016-08-10
Update dotnet-docker-nightly subscription

naamunds:
Use the update-dependencies app to update the CLI versions in the
Dockerfiles.  This will trigger DockerHub builds, so the
build-dotnet-docker-nightly action is no longer needed.

cc: @eerhardt @MichaelSimons

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -30,11 +30,11 @@
       "vsoProject": "DevDiv",
       "buildDefinitionId": 3159
     },
-    // A build definition that will trigger the dotnet-nightly Docker image build in Docker Hub
-    "build-dotnet-docker-nightly": {
+    // A build definition capable of running any .ps1 script in the dotnet-docker-nightly repo
+    "dotnet-docker-nightly-general": {
       "vsoInstance": "mseng.visualstudio.com",
       "vsoProject": "dotnetcore",
-      "buildDefinitionId": 3590
+      "buildDefinitionId": 3784
     }
   },
   "subscriptions": [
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -30,11 +30,11 @@
       "vsoProject": "DevDiv",
       "buildDefinitionId": 3159
     },
-    // A build definition that will trigger the dotnet-nightly Docker image build in Docker Hub
-    "build-dotnet-docker-nightly": {
+    // A build definition capable of running any .ps1 script in the dotnet-docker-nightly repo
+    "dotnet-docker-nightly-general": {
       "vsoInstance": "mseng.visualstudio.com",
       "vsoProject": "dotnetcore",
-      "buildDefinitionId": 3590
```
eerhardt: Are you deleting this old build def?  Why not use the same build def and just update it?

eerhardt:
just one simple question. other than that, LGTM

Maestro/subscriptions.json
MichaelSimons: @eerhardt, yes we are going to delete the old build def.  We had created another build definition already so that we could update the dependencies and test everything out without breaking the existing Maestro hook.  It was a good thing we went down this approach because @naamunds encountered a couple issues that needed to be flushed out.

eerhardt: Sounds good. Let me know when you want this merged.

naamunds: Go ahead and merge it. Thanks.

PR #54, opened 2016-08-16
Rename "UpdateDependencies" target to add "AndSubmitPullRequest".

dagood:
"UpdateDependenices" alone has been changed to only update the files, to better match up target names to actions. Needs to be merged after https://github.com/dotnet/corefx/pull/10868, which takes the change.

/cc @eerhardt


weshaggard:
LGTM

PR #55, opened 2016-08-17
Use new update command in coreclr; add corefx, coreclr subscriptions

dagood:
Accompanies https://github.com/dotnet/coreclr/pull/6664, will merge this after.

`run.cmd build -Project=tests\build.proj -- /t:UpdateDependenciesAndSubmitPullRequest ...` is now used to update coreclr dependencies.

Adds back the CoreCLR and CoreFX dependency updates because now they'll update a single PR when possible instead of sending a flood of them.

@gkhanna79 @wtgodbe @eerhardt 

FYI, @jkotas, that these upgrades will start happening. (I thought you might be interested since you raised the concern about the PR flood a while back.)

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -95,6 +95,25 @@
             "/p:ProjectRepoBranch=master",
             "/p:NotifyGitHubUsers=dotnet/corefx-contrib"
           ]
+        },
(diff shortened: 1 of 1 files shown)


eerhardt:
Looks good to me.

wtgodbe:
Awesome! LGTM

PR #57, opened 2016-08-18
Use UpdateDependencies target for corefx dev/api subscriptions

dagood:
`UpdateDependencies` is the name of the target in the BuildTools version dev/api is using. It later got renamed `UpdateDependenciesAndSubmitPullRequest`, and I'd assumed dev/api took that version.

/cc @weshaggard @joperezr

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -109,7 +109,7 @@
           "ScriptFileName": "build-managed.cmd",
           "Arguments": [
             "--",
-            "/t:UpdateDependenciesAndSubmitPullRequest",
+            "/t:UpdateDependencies",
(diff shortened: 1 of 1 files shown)


weshaggard:
Instead of doing this can we just get the updated buildtools version merged into dev/api?

dagood:
Hmm, could. Do you think we should include the full-version verification config change too? I'm not sure if that will make a future merge from master -> dev/api harder. (I tried to make my buildtools changes compatible with "old" configs, but I didn't verify it.)

weshaggard:
Yes I would merge your entire tools change. Hopefully it shouldn't be too bad as we just finished the earlier merge yesterday.

PR #59, opened 2016-08-25
Add a subscription to the core-setup master branch for official builds.

jhendrixMSFT:
This subscription will kick off an official "pipebuild" of the core-setup
repo per commit into the master branch of the core-setup repo.

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -18,6 +18,12 @@
       "vsoProject": "dotnetcore",
       "buildDefinitionId": 3571
     },
+    // A build definition that will trigger an official build of the core-setup repo
+    "core-setup-pipebuild": {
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
@eerhardt I wasn't sure of the blob path for core-setup master, can you please double-check it?  Other than the scheduled builds in VSTS is there anything else that needs to be removed in order to complete this?

eerhardt:
That path looks right to me.  I'll merge this change if you are ready.

> Other than the scheduled builds in VSTS is there anything else that needs to be removed in order to complete this?

We need to turn off the build triggers in both pipe-build and in the core-setup build definitions.  I can turn the core-setup triggers off, if you want to turn the pipe-build triggers off.

jhendrixMSFT:
@eerhardt I've removed the scheduled pipe build triggers.

PR #60, opened 2016-08-30
Create build-info for ProjectN auto-publish, add dev/api subscription

dagood:
Preparation for auto-publishing of NETNative targeting pack from ProjectN. Create a blank build-info for the auto-publish to fill in and a subscription for corefx dev/api. (I forgot the subscription part when I was talking with you about this, @sedarg.)

This new build-info will include `Microsoft.TargetingPack.Private.NETNative` from ProjectN. I'll be making a manual change to remove this package from `projectk-tfs/Latest_Packages.txt` so auto-upgrade won't hit a conflict. @sedarg is making ProjectK no longer publish that package for future builds.

/cc @eerhardt @sedarg

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -206,6 +206,29 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


eerhardt:
LGTM

PR #61, opened 2016-08-30
Remove projectk-tfs NETNative targeting pack: moving to projectn-tfs

dagood:
Remove projectk-tfs version of the NETNative targeting pack. Follows up on https://github.com/dotnet/versions/pull/60. I thought I would want to order this PR later, but it turns out it's more helpful to do it now.

(This package won't be in projectk-tfs in the next build, this manual change just unblocks some PRs.)

/cc @SedarG @eerhardt


eerhardt:
Shouldn't future builds of `projectk-tfs` just stop publishing this package, which means it will stop showing up?

dagood:
Yes, but in the interest of getting auto-PRs working smoothly sooner I was thinking it would be good to seed this info.

eerhardt:
Will the next build just add it back?

dagood:
@SedarG is working on the ProjectK change that will disable those packages.

eerhardt:
OK, sounds good.  :shipit:

SedarG:
JFYI, I checked in the projectk changes to stop producing and publishing the NetNative targeting pack

PR #62, opened 2016-08-31
Override Run.cmd verbosity to Normal to see logged messages

dagood:
The default verbosity is Minimal, which hides most messages and makes it hard to debug a failure in a Maestro-triggered build.

/cc @eerhardt @weshaggard


dagood:
Went ahead with this so we can start seeing verbose results in the upgrade builds.

eerhardt:
:+1:

PR #63, opened 2016-09-07
Verify payload signatures from webhooks

naamunds:
When Maestro receives an HTTP request, verify that it's from an
authorized source and wasn't corrupted by comparing the payload
signature to the HMAC calculated from the HTTP request's content with
the webhook's secret token as the key.

cc: @eerhardt @bartonjs

### DIFF
--- a/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
+++ b/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
@@ -20,6 +20,7 @@ namespace Microsoft.DotNet.Maestro.WebApi.Controllers
         private EventRegistrationTable _eventService = new EventRegistrationTable();
         private AzureStorageService _storageService = new AzureStorageService();
 
+        [VerifyPayloadSignature]
         public async Task Post(PushWebHookEvent e)
         {
             Trace.TraceInformation($"CommitPushed Started: {e}");
--- a/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Services/VerifyPayloadSignatureAttribute.cs
+++ b/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Services/VerifyPayloadSignatureAttribute.cs
@@ -1,0 +1,128 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Runtime.CompilerServices;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Web.Http.Controllers;
+using System.Web.Http.Filters;
+
+namespace Microsoft.DotNet.Maestro.WebApi.Services
+{
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, Inherited = true)]
+    public class VerifyPayloadSignatureAttribute : FilterAttribute, IAuthorizationFilter
+    {
+        private const string SignatureHeaderName = "X-Hub-Signature";
+
+        public async Task<HttpResponseMessage> ExecuteAuthorizationFilterAsync(
+            HttpActionContext actionContext,
+            CancellationToken cancellationToken,
+            Func<Task<HttpResponseMessage>> continuation)
+        {
+            using (Task<byte[]> getRequestContentTask = actionContext.Request.Content.ReadAsByteArrayAsync())
+            {
+                try
+                {
+                    string payloadSignature = GetPayloadSignature(actionContext.Request);
+                    string computedHash = ComputeHashHexDigest(await getRequestContentTask);
+
+                    if (!CryptographicEquals(payloadSignature, computedHash))
+                    {
+                        Trace.TraceWarning("Signature validation failed.  Payload signature: " +
+                            $"{payloadSignature}; computed hash: {computedHash}");
+
+                        throw new Exception($"The payload signature specified by the {SignatureHeaderName} " +
+                            "header does not match the hash computed from the request.  This could indicate " +
+                            "that the repo's Maestro webhook has an incorrect Secret, the request came from " +
+                            "an unauthorized source, or the request was corrupted.");
+                    }
+                }
+                catch (Exception ex)
+                {
+                    Trace.TraceError(ex.Message);
+
+                    actionContext.Response = actionContext.Request.CreateErrorResponse(
+                        HttpStatusCode.Unauthorized,
+                        ex.Message);
+                }
+            }
+
+            if (actionContext.Response != null)
+            {
+                return actionContext.Response;
+            }
+            else
+            {
+                return await continuation();
+            }
+        }
+
+        private string GetPayloadSignature(HttpRequestMessage request)
+        {
+            IEnumerable<string> headerValues;
+            string requestXHubSignature = string.Empty;
+            if (request.Headers.TryGetValues(SignatureHeaderName, out headerValues))
+            {
+                requestXHubSignature = headerValues.FirstOrDefault();
+            }
+
+            if (string.IsNullOrWhiteSpace(requestXHubSignature) || !requestXHubSignature.StartsWith("sha1="))
+            {
+                throw new Exception($"The HTTP request does not contain a valid {SignatureHeaderName} " +
+                    "header.  Verify that the repo's Maestro webhook is configured with a Secret.");
+            }
+
+            return requestXHubSignature.Substring(5).ToUpper();
+        }
+
+        private string ComputeHashHexDigest(byte[] requestContent)
+        {
+            byte[] hmacKey = Encoding.UTF8.GetBytes(Config.Instance.WebhookSecretToken);
+
+            using (HMACSHA1 hmac = new HMACSHA1(hmacKey))
+            {
+                byte[] computedHash = hmac.ComputeHash(requestContent);
+                return BitConverter.ToString(computedHash).Replace("-", string.Empty);
+            }
+        }
+
+        // Test two strings for equality using a constant-time algorithm such that the execution time varies
+        // only on the length of the data, not the contents thereof, to prevent timing attacks. This method
+        // assumes that the strings are indexable (as all ASCII strings are).
+        [MethodImpl(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
+        private static bool CryptographicEquals(string a, string b)
+        {
+            int result = 0;
+
+            // Short cut if the lengths are not identical
+            if (a.Length != b.Length)
+            {
+                return false;
+            }
+
+            unchecked
+            {
+                // Normally this caching doesn't matter, but with the optimizer off, this nets a non-trivial
+                // speedup.
+                int aLength = a.Length;
+
+                for (int i = 0; i < aLength; i++)
+                    // We use subtraction here instead of XOR because the XOR algorithm gets ever so slightly
+                    // faster as more and more differences pile up. This cannot overflow because the output
+                    // of subtracting two chars is an int.
+                    result = result | (a[i] - b[i]);
+            }
+
+            return (0 == result);
+        }
+    }
+}
--- a/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Microsoft.DotNet.Maestro.WebApi.csproj
+++ b/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Microsoft.DotNet.Maestro.WebApi.csproj
@@ -106,6 +106,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Services\AzureStorageService.cs" />
     <Compile Include="Services\EventRegistrationTable.cs" />
+    <Compile Include="Services\VerifyPayloadSignatureAttribute.cs" />
   </ItemGroup>
   <ItemGroup>
     <Content Include="Global.asax" />
--- a/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Web.config
+++ b/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Web.config
@@ -5,6 +5,7 @@
   -->
 <configuration>
   <appSettings>
+    <add key="WebhookSecretToken" value=""/>
     <add key="VsoUser" value=""/>
     <add key="VsoPassword" value=""/>
     <add key="SubscriptionsUrl" value="https://raw.githubusercontent.com/dotnet/versions/master/Maestro/subscriptions.json"/>
--- a/Maestro/src/Microsoft.DotNet.Maestro/Config.cs
+++ b/Maestro/src/Microsoft.DotNet.Maestro/Config.cs
@@ -14,6 +14,7 @@ namespace Microsoft.DotNet.Maestro
     {
         public static Config Instance { get; } = new Config();
 
+        private Lazy<string> _webhookSecretToken = new Lazy<string>(() => ConfigurationManager.AppSettings[nameof(WebhookSecretToken)]);
         private Lazy<string> _vsoUser = new Lazy<string>(() => ConfigurationManager.AppSettings[nameof(VSoUser)]);
         private Lazy<string> _vsoPassword = new Lazy<string>(() => ConfigurationManager.AppSettings[nameof(VSoPassword)]);
         private Lazy<string> _subscriptionsUrl = new Lazy<string>(() => ConfigurationManager.AppSettings[nameof(SubscriptionsUrl)]);
@@ -21,6 +22,11 @@ namespace Microsoft.DotNet.Maestro
 
         private Config()
         {
+        }
+
+        public string WebhookSecretToken
+        {
+            get { return _webhookSecretToken.Value; }
         }
 
         public string VSoUser


Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
```diff
@@ -82,5 +96,82 @@ public class CommitPushedController : ApiController
...
+                throw new Exception("The HTTP request does not contain a valid X-Hub-Signature header.  Verify that " +
+                    "the repo's Maestro webhook is configured with a Secret.");
+            }
+
+            return requestXHubSignature.Substring(5).ToUpper();
+        }
+
+        private string ComputeHashHexdigest(byte[] requestContent)
+        {
+            byte[] hmacKey = Encoding.UTF8.GetBytes(Config.Instance.WebhookSecretToken);
```
bartonjs: Your HMAC key is UTF8-able?  I'd expect Base64...

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
```diff
@@ -82,5 +96,82 @@ public class CommitPushedController : ApiController
...
+
+            using (HMACSHA1 hmac = new HMACSHA1(hmacKey))
+            {
+                byte[] computedHash = hmac.ComputeHash(requestContent);
+                return BitConverter.ToString(computedHash).Replace("-", string.Empty);
```
bartonjs: If this code gets called a lot, this is pretty expensive (creating the string, then calling Replace on it).  A single-pass hex string build method could be used, or you could read the header hex as bytes instead of make the value bytes be hex.

If this isn't on a perf-critical path it might not be worth the effort, though.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
```diff
@@ -82,5 +96,82 @@ public class CommitPushedController : ApiController
...
+            if (string.IsNullOrWhiteSpace(requestXHubSignature) || !requestXHubSignature.StartsWith("sha1="))
+            {
+                throw new Exception("The HTTP request does not contain a valid X-Hub-Signature header.  Verify that " +
+                    "the repo's Maestro webhook is configured with a Secret.");
+            }
+
+            return requestXHubSignature.Substring(5).ToUpper();
+        }
+
+        private string ComputeHashHexdigest(byte[] requestContent)
```
bartonjs: Possibly want to capitalize the D in Digest?

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
naamunds: The GitHub webhook secret token seems to accept UTF8.  Also, the [OpenSSL::HMAC](http://ruby-doc.org/stdlib-2.1.0/libdoc/openssl/rdoc/OpenSSL/HMAC.html) method that [GitHub recommends using](https://developer.github.com/webhooks/securing/) appears to just take a string.

I've verified that this method of converting to a byte array works with the GitHub webhooks.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
naamunds: I considered leaving the computed hash as a byte array and converting the payload signature from the header to a byte array for the comparison, but opted to convert the computed hash to a string instead so that I could log the hashes in case of a mismatch.

I don't think code is perf-critical because it's just called once per HTTP request to the Maestro endpoint, and there typically aren't that many requests.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
bartonjs: Yeah, things are complex in perl, where everything's just a scalar or not, and C, where strings and byte arrays are both `char*`.  But, if it works, it works...

bartonjs:
CryptographicEquals LGTM.  And since you don't do anything with the message before checking the signature, the usage LGTM, too.  (I apparently never hit submit on this, oops)

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
```diff
@@ -82,5 +96,82 @@ public class CommitPushedController : ApiController
...
+            unchecked
+            {
+                // Normally this caching doesn't matter, but with the optimizer off, this nets a non-trivial
+                // speedup.
+                int aLength = a.Length;
+
+                for (int i = 0; i < aLength; i++)
+                    // We use subtraction here instead of XOR because the XOR algorithm gets ever so
+                    // slightly faster as more and more differences pile up.
+                    // This cannot overflow more than once (and back to 0) because bytes are 1 byte
```
eerhardt: This comment is no longer correct because you are not using bytes, you are using chars.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
```diff
@@ -82,5 +96,82 @@ public class CommitPushedController : ApiController
...
+            using (HMACSHA1 hmac = new HMACSHA1(hmacKey))
+            {
+                byte[] computedHash = hmac.ComputeHash(requestContent);
+                return BitConverter.ToString(computedHash).Replace("-", string.Empty);
+            }
+        }
+
+        // Test two strings for equality using a constant-time algorithm such that the execution time varies
+        // only on the length of the data, not the contents thereof, to prevent timing attacks. This method
+        // assumes that the strings are indexable.
```
eerhardt: Aren't all strings indexable?

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
```diff
@@ -20,12 +24,22 @@ public class CommitPushedController : ApiController
         private EventRegistrationTable _eventService = new EventRegistrationTable();
         private AzureStorageService _storageService = new AzureStorageService();
 
-        public async Task Post(PushWebHookEvent e)
+        public async Task Post()
         {
+            // The PushWebHookEvent object must be manually deserialized instead of using parameter binding 
+            // because the payload signature verification requires access to the request's raw content.  When
+            // using parameter binding, the content is read during the binding and cannot be accessed again due
+            // to Web API treating the request's content as a forward-only stream which can only be read once.
```
eerhardt: I'd like to understand why this GitHub webhook class can read the HttpRequestMessage.Content twice:

[https://github.com/aspnet/WebHooks/blob/master/src/Microsoft.AspNet.WebHooks.Receivers.GitHub/WebHooks/GitHubWebHookReceiver.cs](https://github.com/aspnet/WebHooks/blob/master/src/Microsoft.AspNet.WebHooks.Receivers.GitHub/WebHooks/GitHubWebHookReceiver.cs)

Once here:  [https://github.com/aspnet/WebHooks/blob/master/src/Microsoft.AspNet.WebHooks.Receivers.GitHub/WebHooks/GitHubWebHookReceiver.cs](https://github.com/aspnet/WebHooks/blob/master/src/Microsoft.AspNet.WebHooks.Receivers.GitHub/WebHooks/GitHubWebHookReceiver.cs)#L136

and another time here:  [https://github.com/aspnet/WebHooks/blob/master/src/Microsoft.AspNet.WebHooks.Receivers.GitHub/WebHooks/GitHubWebHookReceiver.cs](https://github.com/aspnet/WebHooks/blob/master/src/Microsoft.AspNet.WebHooks.Receivers.GitHub/WebHooks/GitHubWebHookReceiver.cs)#L74

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
```diff
@@ -82,5 +96,82 @@ public class CommitPushedController : ApiController
 
             Trace.TraceInformation("CommitPushed Complete");
         }
+
+        private void VerifyPayloadSignature(byte[] requestContent)
```
eerhardt: I think it would make sense to move all this logic off into its own class.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
```diff
@@ -20,12 +24,22 @@ public class CommitPushedController : ApiController
...
 
-        public async Task Post(PushWebHookEvent e)
+        public async Task Post()
         {
+            // The PushWebHookEvent object must be manually deserialized instead of using parameter binding 
+            // because the payload signature verification requires access to the request's raw content.  When
+            // using parameter binding, the content is read during the binding and cannot be accessed again due
+            // to Web API treating the request's content as a forward-only stream which can only be read once.
+            byte[] requestContent = await Request.Content.ReadAsByteArrayAsync();
+            PushWebHookEvent e = JsonConvert.DeserializeObject<PushWebHookEvent>(
```
eerhardt: We should probably only deserialize the body after we've verified that the request is OK.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
bartonjs: Something about combining characters, and non-normalized representations of data.  ASCII's all happy, it's things beyond char 127 that get a bit complex.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
naamunds: Fixed.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
eerhardt: This is the same thing the asp.net webhooks repo (which I just learned about this morning) uses:

https://github.com/aspnet/WebHooks/blob/master/src/Microsoft.AspNet.WebHooks.Receivers.GitHub/WebHooks/GitHubWebHookReceiver.cs#L133

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
bartonjs: If you don't need to deserialize to get the signature (or data that was signed) then I agree.

And looking again I see that the content is what's signed, and what's deserialized, so... yeah.  Good catch :)

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
naamunds: I've updated the comment.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
naamunds: I played around with this a bit, and like the aspnet code, I was able to successfully invoke the Request.Content.Read* methods multiple times.  However, when using parameter binding, any attempt to read the content returns empty.  So, it appears that the issue is specific to parameter binding.

I've updated the comment to reflect this finding.

I've also logged https://github.com/aspnet/Mvc/issues/5258 in case it's not the expected ASP.NET behavior.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
naamunds: Updated as suggested.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
naamunds: Good idea. Updated as suggested.

naamunds:
@eerhardt and @bartonjs, I've pushed another commit with updates based on the PR feedback.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Controllers/CommitPushedController.cs
```diff
@@ -20,8 +21,27 @@ public class CommitPushedController : ApiController
...
+
+            try
+            {
+                new RequestValidator(Request, requestContent).VerifyPayloadSignature();
+            }
+            catch (Exception ex)
+            {
+                Trace.TraceError($"An unhandled exception occurred verifying the payload signature : Exception: {ex}");
+
+                throw;
```
bartonjs: What handles this exception?  If, for example, the details get returned to the remote caller you could be telling them the correct signature value for their forgery.  If it is just a generic error (or there's no indication) that's fine.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Services/RequestValidator.cs
```diff
@@ -0,0 +1,108 @@
...
+
+namespace Microsoft.DotNet.Maestro.WebApi.Services
+{
+    public class RequestValidator
+    {
+        private HttpRequestMessage request;
+        private byte[] requestContent;
+        private const string SignatureHeaderName = "X-Hub-Signature";
+
+        public RequestValidator(HttpRequestMessage request, byte[] requestContent)
```
bartonjs: To me the functionality feels like a static method, no instances required; but I don't know if you guys have an anti-static design principle.

eerhardt:
> using System;

missing copyright header

---
Refers to: Maestro/src/Microsoft.DotNet.Maestro.WebApi/Services/VerifyPayloadSignatureAttribute.cs:1 in ff35d76. [](commit_id = ff35d7682dc7105412f1e9b72c84ae51fa829651, deletion_comment = False)

naamunds:
I've updated the verification logic to be in an `IAuthorizationFilter` based on an offline suggestion from @eerhardt.  Incidentally, this allows using parameter binding again rather than needing to manually deserialize the `PushWebHookEvent` object.

naamunds:
> using System;

Fixed.

---
In reply to: [246462347](https://github.com/dotnet/versions/pull/63#issuecomment-246462347) [](ancestors = 246462347)

---
Refers to: Maestro/src/Microsoft.DotNet.Maestro.WebApi/Services/VerifyPayloadSignatureAttribute.cs:1 in ff35d76. [](commit_id = ff35d7682dc7105412f1e9b72c84ae51fa829651, deletion_comment = False)

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Services/VerifyPayloadSignatureAttribute.cs
```diff
@@ -0,0 +1,125 @@
...
+                {
+                    string payloadSignature = GetPayloadSignature(actionContext.Request);
+                    string computedHash = ComputeHashHexDigest(await getRequestContentTask);
+
+                    if (!CryptographicEquals(payloadSignature, computedHash))
+                    {
+                        Trace.TraceWarning($"Payload signature: {payloadSignature}");
+                        Trace.TraceWarning($"Computed hash: {computedHash}");
+
+                        throw new Exception($"The payload signature specified by the {SignatureHeaderName} " +
```
eerhardt: Why not just TraceError here, and then return the Unauthorized response?  Instead of throwing an exception and then catching it, just to return the response.

eerhardt:
:shipit: after responding to my one exception handling comment.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Services/VerifyPayloadSignatureAttribute.cs
naamunds: I wanted to consolidate the logic for tracing an error and creating the error response in one place.  If I do `TraceError` and `CreateErrorResponse` here, then I'd have to do the same thing in another place for the error from `GetPayloadSignature`, right?

eerhardt: OK.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Services/VerifyPayloadSignatureAttribute.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                try
+                {
+                    string payloadSignature = GetPayloadSignature(actionContext.Request);
+                    string computedHash = ComputeHashHexDigest(await getRequestContentTask);
+
+                    if (!CryptographicEquals(payloadSignature, computedHash))
+                    {
+                        Trace.TraceWarning($"Payload signature: {payloadSignature}");
+                        Trace.TraceWarning($"Computed hash: {computedHash}");
```
bartonjs: I don't know how your logging system works; but I'd personally put these in the same message, it avoids trying to re-correlate the message stream.

naamunds: Good idea. Fixed.

bartonjs:
:shipit:

PR #64, opened 2016-09-07
Create release/1.1.0 build-info for corefx, coreclr

dagood:
Prepare empty corefx and coreclr build-infos for release/1.1.0 builds.

/cc @gkhanna79 @weshaggard


weshaggard:
LGTM. We are also going to need a core-setup 1.1

PR #65, opened 2016-09-08
Copy existing build-info into blank files

dagood:
Blank files seem to cause GitHub-side errors (see https://github.com/dotnet/corefx/pull/11269#issuecomment-243868792) so instead of having these files blank, copy over the build-info data from the build-info the new branch depends on. This lets repos take dependencies and subscriptions on the new build-infos without running into that 503 error.

/cc @eerhardt @gkhanna79 @weshaggard


dagood:
Merging this so I can get to work on dependencies.props updates. When a real coreclr build happens this will be replaced with real data.

eerhardt:
:shipit:

PR #66, opened 2016-09-08
Create release/1.1.0 build-info for core-setup

dagood:
Contains data from core-setup's `master`, which `release/1.1.0` branched from. (Only because blank files can trigger a GitHub raw API bug.)

/cc @weshaggard @gkhanna79


eerhardt:
:shipit:

PR #68, opened 2016-09-12
Add subscription to trigger core-setup 1.1.0 builds

dagood:
I said earlier that core-setup builds on commit, but I saw merging a PR that it's only for master. This should add a trigger for release/1.1.0.

I don't think leaving core-setup to be manually queued is a good idea because only one build is possible per commit. Automatically triggering will make sure nobody accidentally queues twice on the same one.

@eerhardt There's no extra per-branch setup in the core-setup repo settings, right?

/cc @gkhanna79 @weshaggard @wtgodbe @ianhays

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -18,11 +18,17 @@
       "vsoProject": "dotnetcore",
       "buildDefinitionId": 3571
     },
-    // A build definition that will trigger an official build of the core-setup repo
-    "core-setup-pipebuild": {
+    // A build definition that will trigger an official build of the core-setup repo on master
+    "core-setup-pipebuild-master": {
       "vsoInstance": "devdiv.visualstudio.com",
       "vsoProject": "DevDiv",
       "buildDefinitionId": 3160
+    },
+    // A build definition that will trigger an official build of the core-setup repo on release/1.1.0
+    "core-setup-pipebuild-release-1.1.0": {
+      "vsoInstance": "devdiv.visualstudio.com",
+      "vsoProject": "DevDiv",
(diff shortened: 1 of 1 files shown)


gkhanna79:
PR LGTM.

> I don't think leaving core-setup to be manually queued is a good idea because only one build is possible per commit.

I agree - as discussed, @wtgodbe will file an issue against @ramarag to fix this.

wtgodbe:
@gkhanna79 I thought we'd discussed filing an issue for moving core-setup's build definitions from MSENG to VSO? I have that here - https://github.com/dotnet/core-setup/issues/256. I don't remember talking about filing an issue against Rama for changing how core-setup should be triggered - can you elaborate on what you want here?

gkhanna79:
Pipeline builds should be triggered in the same way as CoreCLR/CoreFX - time based.

wtgodbe:
Gotcha - I can just go make that change now if you'd like

dagood:
Unless something's changed, it's not that simple: the packages use commit count as the version, so only one build is *possible* per commit until we make significant product changes. A bit of context, although it's for a further end goal: https://github.com/dotnet/core-setup/issues/238.

I think we need to fully pull in buildtools to get the versioning logic to actually get that done, which I understand @ramarag will be doing.

eerhardt:
> Pipeline builds should be triggered in the same way as CoreCLR/CoreFX - time based.

@dagood is correct here.  core-setup can't be built time based, since it uses commit count as the version.

gkhanna79:
@wtgodbe Please file an issue and assign to @ramarag and take it from there.

PR #70, opened 2016-09-13
Add coreclr release/1.1.0 subscriptions

dagood:
* corefx, coreclr -> coreclr (all release/1.1.0)

This depends on merging https://github.com/dotnet/coreclr/pull/7109 which sets up coreclr/release/1.1.0's build-info dependencies.

@gkhanna79 @eerhardt

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -160,6 +160,27 @@
             "/verbosity:Normal"
           ]
         },
+        // This handler will bring the Latest CoreFX release/1.1.0 build into CoreCLR release/1.1.0
+        {
+          "maestroAction": "coreclr-general",
+          "maestroDelay": "00:10:00",
+          "vsoSourceBranch": "release/1.1.0",
+          "ScriptFileName": "run.cmd",
+          "Arguments": [
+            "build",
+            "-Project='tests\\build.proj'",
+            "--",
+            "/t:UpdateDependenciesAndSubmitPullRequest",
+            "/p:GitHubUser=dotnet-bot",
+            "/p:GitHubEmail=dotnet-bot@microsoft.com",
+            "/p:GitHubAuthToken=`$(`$Secrets['DotNetBotGitHubPassword'])",
+            "/p:ProjectRepoOwner=dotnet",
+            "/p:ProjectRepoName=coreclr",
+            "/p:ProjectRepoBranch=release/1.1.0",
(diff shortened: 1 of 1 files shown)


eerhardt:
LGTM

Maestro/subscriptions.json
```diff
@@ -160,6 +160,27 @@
             "/verbosity:Normal"
           ]
         },
+        // This handler will bring the Latest CoreFX release/1.1.0 build into CoreCLR release/1.1.0
```
gkhanna79: I am not sure I follow the comment - CoreCLR does not have a dependency on CoreFX, so why would we need this?

Maestro/subscriptions.json
```diff
@@ -405,6 +426,27 @@
             "/verbosity:Normal"
           ]
         },
+        // This handler will bring the Latest CoreCLR release/1.1.0 build into CoreCLR release/1.1.0
```
gkhanna79: Is this about bringing External CoreCLR dependencies (i.e. the ones from TFS build) into CoreCLR 1.1.0 branch?

Maestro/subscriptions.json
dagood: The CoreCLR tests currently depend on the latest CoreFX build. These aren't just product dependencies but any time versions from one build's packages are used in another repo.

Maestro/subscriptions.json
dagood: This is a true CoreCLR 1.1.0 to CoreCLR 1.1.0 subscription--all it does is change [`CoreClrPackageVersion`](https://github.com/dotnet/coreclr/blob/6bc804a27da5a646c3a1c339e5b653503b379681/dependencies.props#L15) which can optionally be used to run tests.

We decided in the 1.1.0 thread that we'd keep the TFS dependencies at 1.0.0, so there's no coreclr 1.1.0 subscription on projectk-tfs.

dagood:
@gkhanna79 This is ready to merge, but I want to make sure we're on the same page about these auto-updates.

Maestro/subscriptions.json
gkhanna79: Got it. Thanks.

Maestro/subscriptions.json
gkhanna79: Got it.

gkhanna79:
LGTM.

PR #72, opened 2016-09-15
Plug-in wcf to Maestro.

StephenBonikowsky:
@dagood Could you please review.

FYI
@roncain @zhenlan @iamjasonp @mconnew @hongdai

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -47,6 +47,12 @@
       "vsoInstance": "mseng.visualstudio.com",
       "vsoProject": "dotnetcore",
       "buildDefinitionId": 3784
+    },
+    // A build definition capable of running any .ps1 script in the WCF repo. By default, the master branch.
+    "wcf-general": {
+      "vsoInstance": "devdiv.visualstudio.com",
+      "vsoProject": "DevDiv",
+      "buildDefinitionId": 4226
     }
   },
   "subscriptions": [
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -135,6 +141,26 @@
...
           ]
+        },
+        // This handler will bring the Latest CoreFX master build into WCF master
+        {
+          "maestroAction": "wcf-general",
+          "maestroDelay": "00:10:00",
+          "ScriptFileName": "run.cmd",
+          "Arguments": [
+            "build",
+            "-Project='tests\\build.proj'",
```
dagood: This line isn't needed for wcf--it looks like you used the coreclr one as an example, but corefx and coreclr are actually different.

dagood:
Other than that `-Project` line this looks good to me. I checked out the VSTS def and it seemed good.

Maestro/subscriptions.json
StephenBonikowsky: I removed that line in the three places where I added a Wcf handler.

StephenBonikowsky:
FYI, I can't merge, please merge for me once it looks good.

Maestro/subscriptions.json
dagood: I forgot, another difference is that there is no `run build` in the wcf config.json--only the coreclr one has that. It's `build-managed.cmd` in wcf and corefx. (Aka `run build-managed`, although to be consistent the batch file would be better here.)

StephenBonikowsky: Got it.

dagood:
Looks good now, thanks! Merging.

PR #73, opened 2016-09-16
adding a subscription to corefx/master for ProjectN-tfs feed

SedarG:
CoreFx/dev/api has been merged into CoreFX/master and the development will continue on master, rather than dev/api. we need to update the subscription of ProjectN-tfs feed to also update the master branch. 

@dagood , @eerhardt , could you please take a look


dagood:
LGTM, thanks! I'll merge it.

PR #74, opened 2016-09-18
Delay core-setup PRs 10 minutes

dagood:
This lets MyGet get the packages ready: there may be a delay before they can be downloaded which causes restore errors in CI.

The other auto-PRs already do this--I just didn't think to add it when copying over the core-setup auto-PR subscription from master to release/1.1.0. This should fix the restore errors we saw on a few auto-PRs that we worked around by rerunning CI.

@gkhanna79 @wtgodbe


gkhanna79:
LGTM. Merge it.

PR #79, opened 2016-10-05
Add automated validation of subscriptions.json

naamunds:
Building the new validation csproj runs the tests by default by using
the xunit.runner.msbuild NuGet package.

cc: @eerhardt @dagood


naamunds:
To make the product code more accessible from the validation project, I made the private classes inside `EventRegistrationTable` public. After that, `EventRegistrationTable` was simply a wrapper of `SubscriptionsModel` with two simple methods for creating the `SubscriptionsModel` object, so I moved the factory methods to `SubscriptionsModel` and deleted `EventRegistrationTable`.

Maestro/test/Microsoft.DotNet.Maestro.SubscriptionValidation/SubscriptionValidationTests.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.DotNet.Maestro.WebApi.Models;
+using Microsoft.DotNet.Maestro.WebApi.Services;
+using Xunit;
+
+namespace Microsoft.DotNet.Maestro.SubscriptionValidation
+{
+    public class SubscriptionValidationTests
+    {
+        [Fact]
+        public void SubscriptionsJsonShouldBeValidJson()
```
eerhardt: Isn't this test covered by the following 2 tests?

eerhardt:
> using System;

Missing copyright header.

---
Refers to: Maestro/test/Microsoft.DotNet.Maestro.SubscriptionValidation/SubscriptionValidationTests.cs:1 in ae95ab5. [](commit_id = ae95ab59fdf56efa1c6f07b7e3f22193fe67fa6c, deletion_comment = False)

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Models/ActionList.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.DotNet.Maestro.WebApi.Models
+{
+    public class ActionsList
+    {
+        [JsonExtensionData]
+        private IDictionary<string, JToken> ActionNames { get; set; }
```
dagood: Should this be `_actionNames`?

eerhardt:
:shipit:

Maestro/test/Microsoft.DotNet.Maestro.SubscriptionValidation/packages.config
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="xunit" version="2.1.0" targetFramework="net452" />
+  <package id="xunit.abstractions" version="2.0.0" targetFramework="net452" />
+  <package id="xunit.assert" version="2.1.0" targetFramework="net452" />
+  <package id="xunit.core" version="2.1.0" targetFramework="net452" />
+  <package id="xunit.extensibility.core" version="2.1.0" targetFramework="net452" />
+  <package id="xunit.extensibility.execution" version="2.1.0" targetFramework="net452" />
+  <package id="xunit.runner.msbuild" version="2.1.0" targetFramework="net461" />
```
dagood: Is there any value in adding `xunit.runner.visualstudio`? I'm not sure what's usually done.

naamunds: I think it would be valuable for these tests to be run with the build, as is done for other repos like corefx, coreclr, and cli.  Using `xunit.runner.visualstudio` seems to be the simplest way to do that; with it, the following line in the csproj runs the tests:
`<xunit Assemblies="$(OutputPath)\$(AssemblyName).dll" />`

It's also possible to use `xunit.runner.console` in the build by using an Exec task, as described at the bottom of http://xunit.github.io/docs/running-tests-in-msbuild.html, but I don't think that at this time we need the extra flexibility provided by that approach.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Models/ActionList.cs
naamunds: I think that format is only used for fields, not properties.  In https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md, properties are named with Pascal case.

Maestro/test/Microsoft.DotNet.Maestro.SubscriptionValidation/SubscriptionValidationTests.cs
naamunds: Good point.  I've removed the `SubscriptionsJsonShouldBeValidJson` test.

naamunds:
> using System;

Fixed.

---
In reply to: [251693088](https://github.com/dotnet/versions/pull/79#issuecomment-251693088) [](ancestors = 251693088)

---
Refers to: Maestro/test/Microsoft.DotNet.Maestro.SubscriptionValidation/SubscriptionValidationTests.cs:1 in ae95ab5. [](commit_id = ae95ab59fdf56efa1c6f07b7e3f22193fe67fa6c, deletion_comment = False)

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Models/ActionList.cs
dagood: Ah, you're right (and I see now you just moved this class, so it's fine anyway). I don't see properties mentioned in the doc, but it's the default way. I guess I'm just not used to seeing private properties.

Maestro/test/Microsoft.DotNet.Maestro.SubscriptionValidation/packages.config
dagood: Maybe it's just my config, but `xunit.runner.visualstudio` means I can see the tests in the test explorer, and otherwise I can't. Running during the build is fine with me, though.

naamunds: Right; the project would need `xunit.runner.visualstudio` for them to be visible in Test Explorer. We can add that if desired, but I think it would be unnecessary overhead at this point.

PR #80, opened 2016-10-05
Use Latest.txt in corefx release/1.0.0

dagood:
Latest_Packages.txt results in the old auto-update code in corefx release/1.0.0 trying to put the full list of packages into the corefx expected prerelease msbuild property.

I believe the value in the text file is also used in the PR title, causing errors like in https://devdiv.visualstudio.com/DevDiv/_build?buildId=360738 because of the newlines.

Also updates DirPropsVersionElements to `CoreFxExpectedPrerelease` (from `ExternalExpectedPrerelease`), so the correct upgrade happens. It looks like this handler was copy-pasted without enough modifications.


dagood:
FYI @gkhanna79 @weshaggard, if you noticed that 1.0.0 servicing corefx -> corefx auto-updates weren't happening, this was why.

PR #87, opened 2016-10-07
Restore packages in build.cmd

naamunds:
/cc @dagood @eerhardt

### DIFF
--- a/Maestro/build.cmd
+++ b/Maestro/build.cmd
@@ -13,12 +13,22 @@ if not exist "%VS140COMNTOOLS%VsMSBuildCmd.bat" (
   exit /b 1
 )
 
+set cachedNuGet=%LocalAppData%\NuGet\NuGet.exe
 set logFile="%~dp0build.log"
 set solutionPath="%~dp0Microsoft.DotNet.Maestro.sln"
 set buildArgs=/nologo /maxcpucount /nodeReuse:false /v:minimal /clp:Summary /flp:Verbosity=detailed;LogFile=%logFile%;Append %*
 
 REM Add msbuild to the path for the current process.
 call "%VS140COMNTOOLS%\VsMSBuildCmd.bat"
+
+if not exist %cachedNuGet% (
+  echo Downloading latest version of NuGet.exe...
+  if not exist %LocalAppData%\NuGet mkdir %LocalAppData%\NuGet
+  @powershell -NoProfile -ExecutionPolicy unrestricted -Command "$ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest 'https://www.nuget.org/nuget.exe' -OutFile '%cachedNuGet%'"
+)
+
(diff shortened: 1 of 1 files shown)


Maestro/build.cmd
```diff
@@ -13,13 +13,23 @@ if not exist "%VS140COMNTOOLS%VsMSBuildCmd.bat" (
...
 set solutionPath="%~dp0Microsoft.DotNet.Maestro.sln"
 set buildArgs=/nologo /maxcpucount /nodeReuse:false /v:minimal /clp:Summary /flp:Verbosity=detailed;LogFile=%logFile%;Append %*
 
 REM Add msbuild to the path for the current process.
 call "%VS140COMNTOOLS%\VsMSBuildCmd.bat"
 
+if not exist %cachedNuGet% (
+  echo Downloading latest version of NuGet.exe...
+  if not exist %LocalAppData%\NuGet mkdir %LocalAppData%\NuGet
+  @powershell -NoProfile -ExecutionPolicy unrestricted -Command "$ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest 'https://www.nuget.org/nuget.exe' -OutFile '%cachedNuGet%'"
```
eerhardt: Do we care about a specific version?  That way a new version couldn't break us.

Maestro/build.cmd
```diff
@@ -13,13 +13,22 @@ if not exist "%VS140COMNTOOLS%VsMSBuildCmd.bat" (
...
 set logFile="%~dp0build.log"
 set solutionPath="%~dp0Microsoft.DotNet.Maestro.sln"
 set buildArgs=/nologo /maxcpucount /nodeReuse:false /v:minimal /clp:Summary /flp:Verbosity=detailed;LogFile=%logFile%;Append %*
 
 REM Add msbuild to the path for the current process.
 call "%VS140COMNTOOLS%\VsMSBuildCmd.bat"
 
+echo Downloading NuGet.exe v3.4.4.
+if not exist %cachedNuGetDir% mkdir %cachedNuGetDir%
+@powershell -NoProfile -ExecutionPolicy unrestricted -Command "$ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest 'https://dist.nuget.org/win-x86-commandline/v3.4.4/NuGet.exe' -OutFile '%cachedNuGet%'"
```
eerhardt: why are you unconditionally downloading the file now

naamunds: To ensure that the machine has a version of NuGet.exe in this location.  Otherwise it might have a newer version which could break us.

Upon further thought, this behavior could break other apps which are depending on the version of NuGet.exe which is already there, so I think it would be better to put it in `%LocalAppData%\NuGet\3.4.4` instead.  In that case, unconditionally downloading the file isn't necessary to ensure that v3.4.4 is the one we use.

PR #88, opened 2016-10-11
temporarily suspending ProjectNTfs subscription on corefx-master

SedarG:


SedarG:
ProjectN-Tfs feed and its interaction with CoreFX is currently in a weird state. CoreFX cannot be rolled forward unless ProjectN-Tfs feed is being fixed. So with this change I'm temporarily suspending the subscription of the feed from corefx-master.

weshaggard:
Should we instead just comment this out in the corefx repo?

SedarG:
> Should we instead just comment this out in the corefx repo?

That's ok too, but I don't know how to do that.

weshaggard:
@dagood can correct me if I'm wrong but I think you can just comment out this section https://github.com/dotnet/corefx/blob/master/dependencies.props#L44.

dagood:
Commenting it out here only disables the trigger, not the dependency--it would still be upgraded when a different build happens. Commenting this out (or deleting it) would save on no-op builds, but one extra per day isn't a big deal.

dependencies.props is the right place, and you'll also have to comment out https://github.com/dotnet/corefx/blob/63a3f5dd0d0dd8428bb8d411b02a8a2bb805ffd8/dependencies.props#L68-L72 because it references that build-info to do an upgrade/validation.

PR #90, opened 2016-10-14
Refactor subscriptions.json to allow multiple paths

naamunds:
- Allow a Subscription to declare multiple paths that it is subscribed
to.  Whenever any of the files in the paths are changed, the Action the
Subscription references is invoked. This change reduces duplication in
subscriptions.json, because several subscriptions use the same logic for
multiple paths.

- Put arguments for the Action in an "actionArguments" object.  This
removes the need for the built-in Maestro properties (such as "action"
and "delay") to have a prefix, and allows an action argument to have the
same name as a built-in property.

- Detect errors when parsing subscriptions.json by using
JsonPropertyAttribute.Required to ensure all required properties are
present and using JsonSerializerSettings.MissingMemberHandling to ensure
there are no extraneous or misnamed properties.

/cc @dagood @eerhardt


naamunds:
I sorted the subscriptions in subscriptions.json in a separate commit to make it easier to review the file.  For reviewing the changes to subscriptions.json, I'd recommend looking at the two commits individually.

Maestro/subscriptions.json
```diff
@@ -142,54 +90,43 @@
...
-            "/p:ProjectRepoName=wcf",
-            "/p:ProjectRepoBranch=master",
-            "/p:NotifyGitHubUsers=dotnet/wcf-contrib",
-            "/verbosity:Normal"
-          ]
         }
-      ]
+      }
     },
+    // Update dependencies in CoreCLR release/1.0.0
```
dagood: `Update External dependencies in CoreCLR release/1.0.0`?

Maestro/src/Microsoft.DotNet.Maestro/Services/VsoParameterGenerator.cs
```diff
@@ -2,31 +2,29 @@
...
+            actionArguments?.TryGetValue("vsoBuildParameters", out vsoBuildParametersToken);
+
+            if (vsoBuildParametersToken == null)
             {
                 return null;
             }
 
-            JObject parameterObject = new JObject();
-            foreach (KeyValuePair<string, JToken> parameter in vsoParameters)
+            JObject parameterObject = (JObject)vsoBuildParametersToken;
```
dagood: Is it worth throwing a more descriptive error if this isn't a `JObject`? I don't think it'll be a common problem though.

Maestro/subscriptions.json
naamunds: Some update-dependencies subscriptions include dependencies which I think would be considered internal.  For example, the CoreCLR release/1.1.0 subscription updates CoreCLR's dependency on itself.  I don't think it's worth distinguishing between subscriptions which include internal dependencies and subscriptions which don't.

Maestro/src/Microsoft.DotNet.Maestro/Services/VsoParameterGenerator.cs
naamunds: I considered that, but I think the `InvalidCastException` for this case will be clear enough.   If that occurs, a brief look at the code will indicate that a `vsoBuildParameters` element in subscriptions.json isn't an object but needs to be.

Plus, as you mentioned, it shouldn't be a common problem, so I don't think it's worth adding logic for that.

Maestro/subscriptions.json
dagood: I was referring to the property name, to make this similar to the other 1.0.0 subscriptions like `// Update CoreCLR dependencies in CoreFX release/1.0.0`.

It could also be `Update ProjectK TFS dependencies in CoreCLR release/1.0.0`.

naamunds: Ah, I see. Because of how the update-dependencies functionality was implemented in the release 1.0.0 branches, each 1.0.0 subscription can only update one dependency, unlike the 1.1.0 subscriptions.  I agree that it does make sense to specify "ProjectK TFS dependencies". I've made that change.

eerhardt:
Please update [https://github.com/dotnet/versions/blob/master/Maestro/README.md](https://github.com/dotnet/versions/blob/master/Maestro/README.md)

Maestro/subscriptions.json
```diff
@@ -142,54 +90,43 @@
             "/p:NotifyGitHubUsers=dotnet/coreclr-contrib",
```
eerhardt: FYI - Jan just removed these NotifyGitHubUsers.  [https://github.com/dotnet/versions/pull/91](https://github.com/dotnet/versions/pull/91).  Make sure you merge this correctly.

Maestro/src/Microsoft.DotNet.Maestro/Services/VsoParameterGenerator.cs
```diff
@@ -2,31 +2,29 @@
...
 
 namespace Microsoft.DotNet.Maestro.Services
 {
     public static class VsoParameterGenerator
     {
-        public static string GetParameters(IDictionary<string, JToken> vsoParameters, Func<string, bool> filter)
+        public static string GetParameters(IDictionary<string, JToken> actionArguments)
         {
-            if (vsoParameters == null || !vsoParameters.Any())
+            JToken vsoBuildParametersToken = null;
```
eerhardt: (minor) no need to set this to 'null' since the parameter is an 'out' parameter.

eerhardt:
:shipit:

Maestro/src/Microsoft.DotNet.Maestro/Services/VsoParameterGenerator.cs
naamunds: I tried that, but it results in a "use of unassigned local variable" compiler error.

Maestro/subscriptions.json
naamunds: Done.

naamunds:
I've updated the README and merged in Jan's changes to subscription.json.  I'll squash this into the original two commits before merging.

PR #91, opened 2016-10-17
Remove notifications of coreclr/corefx-contrib groups

jkotas:
Remove notifications of coreclr/corefx-contrib groups because of they are not useful


jkotas:
cc @gkhanna79 @stephentoub

PR #100, opened 2016-11-03
Change blank files to "stable" to avoid github raw endpoint bug

dagood:
Retroactively fix the Latest.txt files according to the fix in https://github.com/dotnet/buildtools/pull/1174. This will let dependency verification/updates work better against the latest 1.1.0 versions.

More about a "proper" fix for blank files: https://github.com/dotnet/buildtools/issues/1196

@mellinoe 
/cc @gkhanna79  @weshaggard


gkhanna79:
LGTM.

PR #101, opened 2016-11-07
Remove core-setup release/1.1.0 commit build trigger

dagood:
This will make it so builds don't accidentally get triggered while the 1.1.0 branch is in the fragile stable state.

@gkhanna79


gkhanna79:
LGTM.

PR #103, opened 2016-11-09
Fix projectk/release/1.0.0 build-info to match stable 1.0.x build

dagood:
The auto-publish build successfully pushed the package, but failed to update the versions repo: https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=421214

The versions repo update failure was because auto-publish assumed there would be a prerelease package it could harvest the value from, but there wasn't. I changed the publish script to use `stable` in this case rather than throwing an exception. This matches https://github.com/dotnet/buildtools/pull/1174.

@gkhanna79 @weshaggard

### DIFF
--- a/build-info/dotnet/projectk-tfs/release/1.0.0/Latest.txt
+++ b/build-info/dotnet/projectk-tfs/release/1.0.0/Latest.txt
@@ -1,1 +1,1 @@
-servicing-24709-00
+stable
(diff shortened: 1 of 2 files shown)


gkhanna79:
What happens when we move to 1.0.6-servicing-*?

dagood:
At that point, the script will see the prerelease part of the package version and set `Latest.txt` back to `servicing-xxxxx-xx`. `Latest_Packages.txt` will be updated like normal to the package name/version(s).

gkhanna79:
In that case, LGTM.

PR #105, opened 2016-11-21
Mission control configuration settings

ChadNedzlek:
These values control which releases/branches/repositories are displayed on http://mc.dot.net

You will be able to preview any changes to these files by adding the ?_vcb=your-versions-branch-name
to the URL on that site to load from a branch in this repository prior to checking in any changes.
(Note: This won't work yet because we need to seed this repository first)

### DIFF
--- a/mission-control/sample.txt
+++ b/mission-control/sample.txt
@@ -1,0 +1,84 @@
+{
+	// At any point in this file, a {"$ref": "#/reference/path"} can
+	// be substituted for any value (usually it's used for types).
+
+    "product": "example",
+        // This should be the product name to associate with this
+        // repository on the site. A "product" is a set of things that
+        // all ship together, and all should have the same value. At
+        // the moment, the values for this are "netcore", "netsdk"
+        // and "components"
+    "name": "example-library",
+        // This is an identifier for this repository. Usually this
+        // is the bare repository name from github "corefx", or "coreclr"
+    "displayName": "Example Library",
+        // This is a friendly readable string to be displayed on
+        // the site whenever this repository is referenced
+    "repository": "dotnet/example-library",
+        // The github repository name,
+    "sortOrder": 40,
+        // (Optional) This is used to sort the list of repositories
+        // whenever they are displayed together. If left out, defaults
+        // to 0, and all things with the same order are sorted
+        // alphabetically by name
+    "sources": [
+        // A list of sources of build/test information in Helix
+        // to be displayed as official builds/tests runs
+	{
+		"name": "official/corefx/release/1.0.0/",
+			// Source name from a submitted Helix job, sources
+			// can be seen in the "Private Runs" tab of mc for
+			// personal runs to find the correct anwser for
+			// existing runs,
+		"branch": "release/1.0.0",
+			// The github branch that this source represents
+		"milestone": "1.1.0",
+			// The github milestone this source is tied to
+			// for issues and end dates,
+		"release": "1.1.0",
+			// The product release name is being shipped. This
+			// value is shared among all repositories in the
+			// same "product" (see above) to group builds/tests/etc
+			// into. It is generally related to the milestone,
+			// but that's not required, as long as all repositories
+			// in a product agree on a string,
+		"types": [
+		{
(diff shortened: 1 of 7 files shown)


ChadNedzlek:
@Chrisboh @maririos Here are the configuration files that go with the changes to Mission Control

maririos:
Something that may help people is documentation about the structure of the json. Which sections should they change, which values, etc.
An example of a .json file with the description of each filed may be enough.

ChadNedzlek:
Ooo. An example would be interesting. :-) I'll have to make it so that MC can exclude files (currently it assume _everything_ under this directory is a configuration file). But that shouldn't be too hard.

ChadNedzlek:
Added example file explaining the file and how to preview changes

mission-control/sample.txt
```diff
@@ -0,0 +1,90 @@
...
+        // to 0, and all things with the same order are sorted
+        // alphabetically by name
+    "sources": [
+        // A list of sources of build/test information in Helix
+        // to be displayed as official builds/tests runs
+	{
+		"name": "official/corefx/release/1.0.0/",
+			// Source name from a submitted Helix job, sources
+			// can be seen in the "Private Runs" tab of mc for
+			// personal runs to find the correct anwser for
```
maririos: typo `anwser`

mission-control/sample.txt
```diff
@@ -0,0 +1,90 @@
...
+			// same "product" (see above) to group builds/tests/etc
+			// into. It is generally related to the milestone,
+			// but that's not required, as long as all repositories
+			// in a product agree on a string,
+		"types": [
+		{
+			"name":  "test/functional/cli/",
+				// Type name from a submitted Helix job, sources
+				// can be seen in the "Private Runs" tab of mc for
+				// personal runs to find the correct anwser for
```
maririos: same here

chcosta:
What's the motivation for adding this data to the versions repo and not somewhere more specifically meant for MC?

ChadNedzlek:
I guess I can create an entirely new repository just to contain these handful of configuration files... but they seem to be generally in line with what's going on here: they are files that talk about which branchs/releases/versions of things go where and how they are associated.

Is that something we want to do (and have to manage that repository and disseminate knowledge about it)? Or is there another repository somewhere that makes sense? I'll admit to not being up-to-date on what all the various repositories we have are for. :-)

chcosta:
dotnet/versions specifically relates to the versions produced by each repo.  The version information is utilized in a quite a few different ways, with a gradually increasing focus on utilizing the versions for inter-repo coordination so that we can build the dotnet core product all up.  Maestro is involved because it is the tool by which the versions get updated.  It's not clear to me that MC configuration information should be tied to this repo.  I would think there would be greater value (and less headache), if there was a separate MC / Helix configuration repo for that type of work.  Perhaps an infrastructure repo or something. It doesn't really make sense to me to tie dotnet/versions and MC together.  @eerhardt and @dagood kind of own this repo as far as I know.

markwilkie:
Curious - what data (or format) would be specific to MC that we wouldn't normally tracked in dotnet/versions?

ChadNedzlek:
Some of the important pieces of information that these files track is which branches go with which milestones of a repository, which repositories are shipping as part of the same release vehicle (which we called a "product" in the configuration), and which branches within the repositories of the same vehicle are actually related.

There's also a few mission control specific things in there, though, things like what sorts of builds and test passes do we run for which branches and such.

dagood:
My instinct is that these files are about the builds, not the build *outputs*, so they belong in the product repos like the checked-in build definitions.

Somewhere "which repo/branch combos contain MC config info" would then need to be stored... my initial reaction is that dotnet/versions isn't great for that either, because there might be build configurations that don't produce public assets. (I'm assuming we would still want to see those on MC, but I'm not sure on that.)

ChadNedzlek:
We have found a home for these in another repository. Closing this PR

markwilkie:
Curious where this ended up going?

Chrisboh:
@markwilkie We moved this into the Core-eng repo for now and will adjust its location when we have build from source working.

PR #116, opened 2017-01-17
Remove packages corefx/master no longer builds

dagood:
With the new Last_Build_Packages file functionality, Latest_Packages needs to be cleared when a branch is no longer considered to publish certain packages.

@ericstj This will stop CoreFX from reverting to old packages.

@chcosta This is an interesting implication of this feature which is really only meant for servicing... maybe it would be a good idea to only turn it on for release branches?


chcosta:
@dagood Can you link to the "new Last_Build_Packages" file functionality?

dagood:
Here's the issue about the feature I'm talking about (maybe the way I'm referring to it isn't clear): https://github.com/dotnet/buildtools/issues/1117

chcosta:
I think my confusion may be because I don't recall the implementation details for the Last_Build_Packages file.  I think you're saying that it's populated based on whatever packages a build produces.  So, if we were to clear the file, then, in our servicing branches, it would only get re-populated with whatever package we happened to be servicing at the time and not with the full set of packages that branch *could* produce.  

I'm not certain how turning on the "Last_Build_Packages" feature only for servicing would solve the problem, wouldn't that mean we only saw the problem in servicing? 

The simple solution is to add a switch for servicing saying don't clear the package list.  My preference is to have something that just works without needing to manually set said switch, so maybe key off of the prerelease moniker and if it's "servicing", don't clear the list, anything else, clear the list.

dagood:
Here's what enabled vs. disabled mean for this feature with more granularity:

Disabled:
 * `Latest` is cleared and set to the current build's outputs.

Enabled:
 * `Last_Build` is cleared and set to the current build's outputs.
 * `Latest` is merged with the current build's outputs, where the current build wins.

In this PR I effectively cleared `Latest` in a branch where the feature is enabled, so I trimmed out the old packages. Disabling the feature then running a build would have done the same thing (although `Last_Build` wouldn't be updated).

Leaving this disabled on master means package removals will automatically update the versions repo. Not having a `Last_Build` file makes the naming convention a little inconsistent, but we could change to publish both files, or delete `Last_Build`.

Enabling this on servicing allows selective package building to correctly update the versions repo. I think the `Last_Build` file is just a way to see what was built without looking at diffs.

I agree that keying off the prerelease moniker is a good idea. (Just need to make sure what we key off stays `servicing` in the final release build--so no examining packages.)

PR #117, opened 2017-01-18
Initialize corert.

crummel:
Add files for CoreRT build to update.


crummel:
@dagood if you don't mind taking a look.

dagood:
If CoreRT uses a BuildTools/VersionTools with https://github.com/dotnet/buildtools/pull/1287, this shouldn't be needed. No harm in merging this just to unblock things, though, so I'll do that.

PR #124, opened 2017-02-06
Add standard repo to the auto-update PRs for corefx and core-setup

weshaggard:
@eerhardt

FYI @ericstj @bryanar


eerhardt:
Is this enough?  Does the update script know to look for new netstandard versions and to update the correct places in the repo?

/cc @dagood

weshaggard:
The update scripts in the repo's may need some adjusting as well but this should get the new packages flowing and we can update corefx and core-setup update scripts accordingly.

dagood:
Here's the related core-setup PR--it needs `standardVersionUrl` plugged in: https://github.com/dotnet/core-setup/pull/1347. core-setup and corefx's systems should both handle NOPs fine until they get updated.

eerhardt:
From looking at https://github.com/dotnet/core-setup/pull/1347, someone will need to update core-setup to point that code to https://github.com/dotnet/versions/tree/master/build-info/dotnet/standard/master, since no one is passing in `STANDARD_VERSION_URL` env var.

PR #127, opened 2017-02-14
Fix CLI trigger path

eerhardt:
I had an incorrect URL for the trigger path in the CLI subscription.

@MichaelSimons @dagood @naamunds


eerhardt:
I'm thinking we could have a test here that validates the triggerPath is "well-formed".  I opened #128 for it.

I'm pretty sure the way I got this incorrect path was clicking through the path in the GitHub UI, and then copying the URL out - which has "tree" in the URL.

PR #131, opened 2017-02-22
Revert corefx/release/1.1.0 build-info update

dagood:
This reverts commits 63eeda89c687d359d42cb6f824c2a2fcaf459fcf and 7a70e49511fdf08fb99b98d92dbdcbd0d75c5952, created when an unintentional corefx 1.1.0 build happened. I put the packages from the earlier build back in place on MyGet, and this PR makes the dotnet/versions repo consistent with the rollback.

See https://github.com/dotnet/coreclr/pull/9708

@gkhanna79 @MattGal

### DIFF
--- a/build-info/dotnet/corefx/release/1.1.0/Latest.txt
+++ b/build-info/dotnet/corefx/release/1.1.0/Latest.txt
@@ -1,1 +1,1 @@
-servicing-25021-01
+servicing-25009-02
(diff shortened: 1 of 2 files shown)


gkhanna79:
Thanks a lot @dagood. The change looks good to me.

What about packages in the Azure blob - will they need to be reverted/rolled back as well?

CC @leecow.

dagood:
The CoreCLR packages shouldn't be in any stable-versioned container, so nothing to fix there. (The only container involved in the product bits of a CoreFX/CoreCLR-style build is the intermediate storage location between the build legs and the publish leg.)

PR #136, opened 2017-03-07
Prune packages no longer produced in corefx/master

dagood:
I noticed some packages weren't updated in https://github.com/dotnet/versions/commit/ca917e7e2e1db8d17e279c045758a1abc62a7210#diff-b64b0edcf4b46fafc86baec415ff9201, indicating that CoreFX no longer builds them. Servicing is the reason to leave packages from old builds in the file, and this is a master build where that doesn't apply.

Is it useful to be able to see when packages stop being produced, @weshaggard, or should master branch builds always remove these old entries?


weshaggard:
>Is it useful to be able to see when packages stop being produced, @weshaggard, or should master branch builds always remove these old entries?

I think that should be configurable per branch. For branches where we are filtering the packages in packages.builds we should keep the "merge" option for this but for branches where we don't filter, like master, then it should get reset each time.

PR #139, opened 2017-03-24
Initializing dotnet/commonrefactor

StephenBonikowsky:


StephenBonikowsky:
@dagood is this correct?

dagood:
Not quite, the paths are case sensitive and include the repo name, like `build-info/dotnet/wcf/CommonRefactor/Latest.txt` The files being empty with no newline is good, though.

PR #147, opened 2017-04-25
Add CLI release/2.0.0 update-dependencies.

eerhardt:
@dagood @livarcocc


eerhardt:
> This made me notice that the new CLI branch is called "release" rather than "rel" this time, interesting. 

CLI is explicitly trying to conform to the rest of the dotnet repos where it makes sense.

PR #152, opened 2017-05-05
Adding placeholder file to core-setup so that we can use the UpdateDependencies task in corefx

joperezr:
PTAL: @dagood 

Adding place holder Latest.txt file to core-setup master so that we can get latest package versions from corefx repo.

### DIFF
--- a/build-info/dotnet/core-setup/master/Latest.txt
+++ b/build-info/dotnet/core-setup/master/Latest.txt
@@ -1,0 +1,1 @@
+placeholder


dagood:
LGTM.

https://github.com/dotnet/core-setup/pull/2292 will start updating this file with the real values since it uses UpdatePublishedVersions, but putting a placeholder in for now unblocks some auto-PR work in CoreFX. Merging.

FYI @karajas @eerhardt @weshaggard

PR #153, opened 2017-05-09
Add definition to update core-setup master

karajas:
This is allow core-setup to receive auto-update PRs.

/cc @dagood @chcosta @weshaggard

PR #154, opened 2017-05-10
Set core-setup master auto-update branch

dagood:
The core-setup-general definition defaults to release/1.0.0, so master must be specified.

I didn't think of this when reviewing https://github.com/dotnet/versions/pull/153.

/cc @karajas @chcosta @weshaggard


dagood:
I'm going ahead with the merge so we don't miss opportunities to exercise this auto-PR flow.

karajas:
Thanks Davis!

PR #155, opened 2017-05-10
Notify dotnet/core-setup-contrib on master auto-PR

dagood:
@dotnet/core-setup-contrib 

For CoreCLR there are separate `dotnet/coreclr-contrib` and `dotnet/coreclr-auto-update-notify` groups. We could do that for core-setup as well. Let me know and I can create a `dotnet/core-setup-auto-update-notify` team.

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -340,6 +340,7 @@
             "/p:ProjectRepoOwner=dotnet",
             "/p:ProjectRepoName=core-setup",
             "/p:ProjectRepoBranch=master",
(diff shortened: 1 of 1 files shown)


dagood:
@karajas Creating `dotnet/core-setup-auto-update-notify` would be the best way to do that, if desired. Escaping the `NotifyGitHubUsers` property properly is surprisingly difficult because of the various layers it passes through. :smile: Also, there have been complaints about the spamminess of auto-PR notifications before, so it's good to let people leave the group if they don't need them.

PR #157, opened 2017-05-18
Update docker subscription to pickup cli master builds

MichaelSimons:

PR #159, opened 2017-05-19
Add a NuGet.Config to ensure nuget.org feed used

dagood:
CI has sometimes been unable to restore packages because it is only looking at local sources. Add a NuGet.Config file to make sure we can acquire the packages, no matter what the machine state is.

See https://github.com/dotnet/versions/pull/157, https://github.com/dotnet/versions/pull/158

### DIFF
--- a/Maestro/NuGet.Config
+++ b/Maestro/NuGet.Config
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+  </packageSources>
+</configuration>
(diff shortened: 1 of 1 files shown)


Maestro/NuGet.Config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
eerhardt: Should we have a <clear /> here so that we only get packages from the official feed?

dagood: Guessing `<clear />` disappeared as html. I'll add it--local sources wouldn't be bad, but we might as well protect against worse problems.

PR #164, opened 2017-05-26
Split dev3 pushes from projectn-tfs-testilc/master

dagood:
Moves existing dev3 branch info into a new "branch", and reverts TestILC master to what it was before dev3 package versions overwrote it.


dagood:
Yeah, 25325-00 is the latest on the feed.

PR #165, opened 2017-06-01
Clean up old packages we no longer produce from package list

weshaggard:
PTAL @dagood @gkhanna79


weshaggard:
>You copied from Last_Build_Packages.txt to Latest_Packages.txt to produce these, right? LGTM.

Yes

PR #166, opened 2017-06-07
Adding a new subscription fro dev/defaultintf

jcagme:
Enabling auto-upgrades for core-setup's dev/defaultintf branch

/cc: @dagood

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -413,9 +413,7 @@
     // Update dependencies in core-setup dev/defaultintf
     {
       "triggerPaths": [
-        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/dev/defaultintf/Latest.txt",
         "https://github.com/dotnet/versions/blob/master/build-info/dotnet/coreclr/dev/defaultintf/Latest.txt",
-        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/standard/dev/defaultintf/Latest.txt"
       ],
       "action": "core-setup-general",
       "delay": "00:10:00",
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -410,6 +410,34 @@
         }
       }
     },
+    // Update dependencies in core-setup dev/defaultintf
+    {
+      "triggerPaths": [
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -410,6 +410,34 @@
         }
       }
     },
+    // Update dependencies in core-setup dev/defaultintf
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/dev/defaultintf/Latest.txt",
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/coreclr/dev/defaultintf/Latest.txt",
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/standard/dev/defaultintf/Latest.txt"
```
dagood: Will there be `dev/defaultinf` branches building for CoreFX and Standard, too?

Maestro/subscriptions.json
```diff
@@ -410,6 +410,34 @@
...
+          "Arguments": [
+            "--",
+            "/t:UpdateDependenciesAndSubmitPullRequest",
+            "/p:GitHubUser=dotnet-bot",
+            "/p:GitHubEmail=dotnet-bot@microsoft.com",
+            "/p:GitHubAuthToken=`$(`$Secrets['DotNetBotGitHubPassword'])",
+            "/p:ProjectRepoOwner=dotnet",
+            "/p:ProjectRepoName=core-setup",
+            "/p:ProjectRepoBranch=dev/defaultintf",
+            "/p:NotifyGitHubUsers=dotnet/core-setup-contrib",
```
dagood: Might be better to not notify this group for now, to avoid noise while this branch gets off the ground. (Just remove this line.)

Maestro/subscriptions.json
jcagme: Good catch, I don't thing they will at this point

dagood:
FYI @ellismg

ellismg:
Thanks!  I think Yi may end up needed to making a branch in corefx as well (so he can expose stuff from the reference assemblies), so setting up maestro for corefx/master's packages may end up being work we have to change.

PR #173, opened 2017-06-22
Updating build-info/dotnet/cli/release/2.0.0 for preview2-006497

MichaelSimons:


MichaelSimons:
@dagood, this can be merged independently.  ASP.NET Core wants updated Docker images now.

PR #178, opened 2017-08-07
Add corefx Master mirror build

ChadNedzlek:
This should fix https://github.com/dotnet/core-eng/issues/1084, where we are having rate limiting issues with some of the bots because of how VSO sets up the CI trigger


ChadNedzlek:
@dagood This looks reasonable? I don't want to merge this until I remove the CI trigger from the job itself.

Maestro/subscriptions.json
```diff
@@ -299,6 +305,13 @@
         }
       }
     },
+    // Mirror corefx changes into VSO
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/**/*",
+      ],
+      "action": "corefx-mirror"
+    },
```
dagood: This will only work with the master branch trigger, not any of the other branches. Looks fine as an incremental change, but make sure only to disable the master mirror build def trigger. With how Maestro currently works, you'll need a subscription entry for each branch, specifying `vsoSourceBranch`.

Maestro/subscriptions.json
```diff
@@ -299,6 +305,13 @@
         }
       }
     },
+    // Mirror corefx changes into VSO
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/**/*",
```
dagood: This points to changes in the versions repo, and should point to the corefx repo instead: `https://github.com/dotnet/corefx/blob/master/**/*`

Maestro/subscriptions.json
ChadNedzlek: Added targets for the other branches that actually exist (I didn't add dev/api or dev/eng, since those branches don't exist in corefx anymore, so there is no reason to add triggers that will never fire)

Maestro/subscriptions.json
ChadNedzlek: Oops, didn't notice the trigger had both versions and corefx in it. Updated all the paths.

Maestro/subscriptions.json
dagood: Set of branches looks good to me.

ChadNedzlek:
Web hook has been created /cc @dagood @mmitche

weshaggard:
@ChadNedzlek Can you guys let me know why we switched to this type of trigger for the corefx mirror?

PR #179, opened 2017-08-08
Update subscriptions for docker sdk 1.1.0 dependency

MichaelSimons:


Maestro/subscriptions.json
```diff
@@ -605,10 +605,10 @@
         }
       }
     },
-    // Update cli rel/1.0.1 dependencies in dotnet-docker-nightly master
+    // Update cli rel/1.1.0 dependencies in dotnet-docker-nightly master
     {
       "triggerPaths": [
-        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/cli/rel/1.0.1/Latest_Packages.txt"
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/cli/rel/1.1.0/Latest_Packages.txt"
```
dagood: It looks like the contents of the file are thrashing quite a bit.

MichaelSimons:
Agreed the build-info is not very healthy.  Will push CLI to fix their end.  Regardless this is the change needed to auto-update the Docker repo.

PR #183, opened 2017-08-16
Add branch parameter to docker subscription

MichaelSimons:
vsoSourceBranch was missing from docker build subscription.  Also made a change to the docker version info to trigger an end-to-end test.


Maestro/subscriptions.json
```diff
@@ -640,7 +640,10 @@
         "https://github.com/dotnet/versions/blob/master/build-info/docker/debian/jessie.txt",
         "https://github.com/dotnet/versions/blob/master/build-info/docker/debian/stretch.txt"
       ],
-      "action": "dotnet-docker-nightly-pipebuild"
+      "action": "dotnet-docker-nightly-pipebuild",
+      "actionArguments": {
+        "vsoSourceBranch": "master"
```
dagood: Interesting... do you know why defaulting wasn't working?

MichaelSimons: What defaulting are you referring to?  The $(Build.SourceBranch) variable is empty in the dotnet-docker-nightly-pipebuild which causes pipebuild to throw and ArgEx.  Maestro has no defaulting and it appears vso doesn't either at the variable level.  Going out on a limb here that this may not cause an issue for the built in `Get Sources` because it may have defaulting logic.

dagood: Yeah, I assumed the VSTS `Get Sources` default (which we rely on for other subscriptions) would be applied uniformly in the build, but I guess not to variables.

PR #197, opened 2017-10-26
Set vsosourcebranch for standard subscrilption

chcosta:
@dagood PTAL

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -753,8 +753,8 @@
       "action": "standard-general",
       "delay": "00:10:00",
       "actionArguments": {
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -754,6 +754,7 @@
       "delay": "00:10:00",
       "actionArguments": {
         "vsoBuildParameters": {
+          "vsoSourceBranch": "master",
```
dagood: This should actually be an `actionArguments` property:

https://github.com/dotnet/versions/blob/b32284828bce0b1e357abdc8c29b78f8924d6013/Maestro/subscriptions.json#L131-L134

Issue #205, opened 2018-01-09
Add a way to pass dotnet/versions commit hash into the triggered VSTS build


dagood:
This can be used by the triggered def to make sure each file change is only processed once. Right now, if two changes come in quickly, both triggered builds will see the same latest dotnet/versions state. If the builds know the dotnet/versions commits that triggered them, they can be sure sure they're the only build triggered for that specific commit.

PR #187, opened 2017-08-22
Add support for passing trigger information to the handlers

weshaggard:
PTAL @eerhardt @dagood 

This will add some well known arguments `<trigger-repo>`, etc to get information about what triggered the run. I brought this together yesterday while it was top of mind but I will likely finish it after the current servicing releases.

PR #206, opened 2018-01-10
Add a parameter substitution for commit id

dagood:
https://github.com/dotnet/versions/issues/205

Builds on the substitution feature added in https://github.com/dotnet/versions/pull/187

I see `id` defined in real webhook payloads, and also defined here: https://developer.github.com/v3/activity/events/types/#webhook-payload-example-26


dagood:
I'm going through the deployment now.

Issue #110, opened 2016-12-20
Add "triggerRepository" property that doesn't require files to change


dagood:
In ~~Core-Setup~~ CLI, a commit is required before running a new build. Sometimes a simple retry is desired, so we need to make a dummy commit. The cleanest possible dummy commit is one that changes no files, which means Maestro currently can't detect it and trigger the build. Adding a feature to run an action even if no files changed would allow us to use no-op dummy commits.

dagood:
Additional scenario: mirroring a newly created release branch before any changes are pushed to the branch.

PR #210, opened 2018-01-23
Switch to a general purpose vsts to github mirror build definition

weshaggard:
cc @mmitche @dagood 

I've created a general purpose VSTS build definition for mirroring changes. I will test it out on corefx and once it appears to be working OK I will switch others to using it as well.

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -6,11 +6,11 @@
       "vsoProject": "DevDiv",
       "buildDefinitionId": 1084
     },
-    // A 'build' that mirrors corefx github into VSO
-    "corefx-mirror": {
+    // A 'build' that mirrors changes from github into VSTS
+    "github-vsts-mirror": {
       "vsoInstance": "devdiv.visualstudio.com",
       "vsoProject": "DevDiv",
-      "buildDefinitionId": 1119
+      "buildDefinitionId": 1308884
     },
     // A build definition capable of running any .ps1 script in the CoreFxLab repo. By default, the master branch.
     "corefxlab-general": {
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -6,11 +6,11 @@
...
       "buildDefinitionId": 1084
     },
-    // A 'build' that mirrors corefx github into VSO
-    "corefx-mirror": {
+    // A 'build' that mirrors changes from github into VSTS
+    "github-vsts-mirror": {
       "vsoInstance": "devdiv.visualstudio.com",
       "vsoProject": "DevDiv",
-      "buildDefinitionId": 1119
+      "buildDefinitionId": 1308884
```
dagood: This is a build id, not a definition id--should be `8289`.

dagood:
This won't work for empty commits before we do https://github.com/dotnet/versions/issues/110. Could become a problem for Core-Setup `release/1*` and CLI due to people adding empty commits to trigger the CI-style official builds that rely on commit depth. (Might be aware, but just in case.)

Maestro/subscriptions.json
weshaggard: Oops thanks. I thought that number looked a little high.

weshaggard:
For mirroring purposes I don't think empty commits are very interesting, so I think we are good for this case.

dagood:
I have errors in mind that can actually happen if an empty commit isn't mirrored: if a build triggers on a push to VSTS, it wouldn't trigger anymore on empty commits, and we have assumptions that the latest commit on github is available in VSTS.

But if this hits someone, it's easy to tell them they have to make a meaningless change rather than empty commit.

weshaggard:
I see yes that makes sense. If they are depending on the empty commits triggering a build then that will be an issue. I think it is OK to for a dummy change as you say though.

PR #218, opened 2018-01-29
Point release/2.1 publish at real locations

dagood:
Moves publish away from the `-test` locations.

We shouldn't merge this until we know the manifests are ok and won't overwrite blobs built by `master`. aspnet requires some build-to-build overwriting because it includes its "latest" files in the manifest, so we can't just disable overwrites.


dagood:
I'm merging this to get final publish going to real locations: the blobs I was worried about have been renamed to release/2.1 and manifests have been looking good. We need these publishes to go to the real locations for the breaking change partial publish flow to run smoothly.

FYI @mmitche

PR #222, opened 2018-01-31
 Add Core-Setup release/2.1 BuildTools subscription

dagood:
When https://github.com/dotnet/core-setup/pull/3655 is merged, we can merge this to get BuildTools 2.1 update for Core-Setup 2.1.

I also cleaned up the buildtools/2.1 build-info, which would be confusing.


Maestro/subscriptions.json
```diff
@@ -530,6 +530,32 @@
         }
       }
     },
+    // Update dependencies in core-setup release/2.1
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/buildtools/release/2.1/Latest.txt"
```
eerhardt: Any reason not to add corefx and coreclr triggers here?

/cc @weshaggard

eerhardt:
It looks like https://github.com/dotnet/versions/tree/master/build-info/dotnet/buildtools/2.1 is still there.

Maestro/subscriptions.json
dagood: Doesn't hurt anything to trigger on paths that don't exist, but it could be confusing since Core-Setup dependencies.props can only take BuildTools upgrades. (Because that will 404 if it's configured to upgrade on a build-info that doesn't exist, and updaters would be confused about packages that don't exist, too.)

Also not sure yet if we're going to uptake prodcon 2.1 builds through the individual build-infos: https://github.com/dotnet/core-eng/issues/2435. Doesn't necessarily impact what we do now, though.

PR #223, opened 2018-02-02
Enable Copy to Latest for 2.1

dagood:
Start copying 2.1 Core-Setup and CLI builds to the `release/2.1` channel. This gives us the stable links on `dotnetcli`/`dotnetclichecksums` that are used in the Core-Setup and CLI repo readmes. (Those readmes aren't suitable to auto-upgrade, unlike the readme on dotnet/versions.)


dagood:
Sorry for the churn, realized these properties have to be set on `master` too or else the publish step will break. It won't interfere with isolated `master` builds because the default publish account is `dotnet-test` rather than `dotnet`.

PR #232, opened 2018-02-16
Add Core* auto-update triggers for release/2.1

dagood:
https://github.com/dotnet/coreclr/pull/16405
https://github.com/dotnet/corefx/pull/27180
https://github.com/dotnet/core-setup/pull/3698

https://github.com/dotnet/core-eng/issues/2435

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -243,6 +243,36 @@
         }
       }
     },
+    // Update dependencies in CoreCLR release/2.1
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/release/2.1/Latest.txt",
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/coreclr/release/2.1/Latest.txt",
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/buildtools/release/2.1/Latest.txt",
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -243,6 +243,36 @@
         }
       }
     },
+    // Update dependencies in CoreCLR release/2.1
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/release/2.1/Latest.txt",
```
weshaggard: Is this correct? I was working under the assumption that we would be triggering the auto updates based on https://github.com/dotnet/versions/blob/master/build-info/dotnet/product/cli/release/2.1/build.xml? 

Is the final publish also still publishing to the individual repos? And if so do we really need to do that?

dagood: That's the ultimate goal (https://github.com/dotnet/core-eng/issues/2675) but we decided the MVP is to split up the manifest into the old format for each repo and continue using that flow with essentially unchanged configs checked into the Core repos for 2.1-preview1.

But even with the MVP, these *should* probably trigger on `packages.semaphore` for the product built bits to avoid triggering each subscription multiple times for the same build. (The build-infos are generated as part of the package publish step, since the packages listed will only be available after then anyway.) I'll make this change.

dagood: What's missing for getting the info from the manifest directly is a way to configure VersionTools through MSBuild to fetch the manifest and use packageversions from that file, and changing the repos to use that instead of the current way.

PR #243, opened 2018-02-23
Subscribe aspnet-docker nightly to updates of release/2.1

natemcmaster:
Resolves https://github.com/aspnet/aspnet-docker/issues/256

cc @dagood @MichaelSimons @ryanbrandenburg


natemcmaster:
Updated to match changes in build definition

PR #248, opened 2018-02-23
Use mirror which can handle private repos

chcosta:
Arcade is still a private repo for the moment,  I needed to clone the mirror definition to handle an authenticated GitHub url clone.  

/cc @dagood @weshaggard


weshaggard:
Is this something we should add as a parameter to the general mirror definition instead of creating a new definition?

PR #256, opened 2018-03-02
Add new repo to be mirrored

eerhardt:


Maestro/subscriptions.json
```diff
@@ -502,7 +502,8 @@
     // Mirror private github changes to vsts
     {
       "triggerPaths": [
-        "https://github.com/dotnet/arcade/blob/master/**/*"
+        "https://github.com/dotnet/arcade/blob/master/**/*",
```
weshaggard: Any reason it has to be in the private mirror? @chcosta I believe is planning to remove this one.

eerhardt: because it isn't a public repo.  Looking at the build defs, the other one doesn't use any permissions when cloning the GitHub repo.

chcosta: When I fold the private mirror back into the standard mirror, I'll transition both of these to use that mirror as well.  I haven't had time to do that recently.

PR #260, opened 2018-03-09
The package was pushed, but dotnet/versions wasn't updated.

nattress:
We updated the ProjectN targeting pack to include a modified System.Private.CoreLib that includes @stephentoub 's recent public API change. This had to be done manually (not in nightly build) since we cannot merge CoreRT into TFS until we first get a UAP build out of CoreFX.


nattress:
@dagood PTAL

PR #267, opened 2018-03-13
Switch WCF release/2.0.0 to use dotnet-maestro-bot

dagood:
This is the only place `dotnet-bot` is used in a subscription. Moving to the preferred `dotnet-maestro-bot` account lets us simplify the definitions and reduce the permissions that auto-PR generation code has access to.

The only visible change will be that WCF release/2.0.0 auto-PRs start coming from @dotnet-maestro-bot, like they do for the other WCF branches already.

Related to https://github.com/dotnet/core-eng/issues/2382

FYI @StephenBonikowsky

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -873,9 +873,9 @@
           "Arguments": [
             "--",
             "/t:UpdateDependenciesAndSubmitPullRequest",
-            "/p:GitHubUser=dotnet-bot",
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
Thanks @dagood although we don't yet have auto-PRs hooked-up yet for WCF release/2.0.0. We were not able to move that branch to the buildtools v2 that was produced and had to continue using the buildtools for 1.1.0.

PR #268, opened 2018-03-14
Have the subscription for WCF 2.0.0 use the new UpdateDep script.

StephenBonikowsky:
The WCF branch 2.0.0 does not use buildtools v2 so using the workaround that was used for 1.1.0 regarding the TLS 1.2 requirement.
Related to dotnet/wcf#2676

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -869,7 +869,7 @@
       "actionArguments": {
         "vsoSourceBranch": "release/2.0.0",
         "vsoBuildParameters": {
-          "ScriptFileName": "build-managed.cmd",
+          "ScriptFileName": "UpdateDependencies.ps1",
           "Arguments": [
             "--",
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
@dagood can you review?

Maestro/subscriptions.json
```diff
@@ -869,7 +869,7 @@
       "actionArguments": {
         "vsoSourceBranch": "release/2.0.0",
         "vsoBuildParameters": {
-          "ScriptFileName": "build-managed.cmd",
+          "ScriptFileName": "UpdateDependencies.ps1",
           "Arguments": [
             "--",
             "/t:UpdateDependenciesAndSubmitPullRequest",
```
dagood: The `--` and `/t:UpdateDependenciesAndSubmitPullRequest` lines need to be removed, too. `--` isn't needed in this flow, and the target is already included in the script.

PR #271, opened 2018-03-16
Force Maestro to submit a PR for uwp-setup release

nattress:
Testing the new branch auto-PR flow but don't have a newer set of packages


nattress:
@dagood PTAL. I want Maestro to submit a PR to update the old packages in uwp-setup rel/6.1. Zach tells me tickling Latest.txt is the way to go about that.

PR #272, opened 2018-03-20
Update source-build from ProdCon 2.1 builds

dagood:
A subscription to submit an auto-update PR to source-build whenever a 2.1 ProdCon build finishes. The PR updates the submodules based on the commits in the orchestrated build manifest. [New source-build-general VSTS def](https://devdiv.visualstudio.com/DevDiv/Default/_build/index?context=allDefinitions&path=\DotNet\Maestro&definitionId=8760&_a=completed).

Waiting for https://github.com/dotnet/source-build/pull/371 to add the prodcon submodule updaters to source-build.

https://github.com/dotnet/source-build/issues/342

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -90,6 +90,12 @@
       "vsoProject": "DevDiv",
       "buildDefinitionId": 7263
     },
+    // A build definition capable of running any .ps1 script in the source-build repo
+    "source-build-general": {
+      "vsoInstance": "devdiv.visualstudio.com",
+      "vsoProject": "DevDiv",
+      "buildDefinitionId": 8760
+    },
     // A build definition capable of running any .ps1 script in the Standard repo. By default, the master branch.
     "standard-general": {
       "vsoInstance": "devdiv.visualstudio.com",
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -1175,6 +1181,29 @@
...
+          "ScriptFileName": "build.cmd",
+          "Arguments": [
+            "/t:UpdateToRemoteDependencies",
+            "/t:SubmitPullRequestIfChanged",
+            "/p:GitHubUser=dotnet-maestro-bot",
+            "/p:GitHubEmail=dotnet-maestro-bot@microsoft.com",
+            "/p:GitHubAuthToken=`$(`$Secrets['DotNetMaestroBotGitHubToken'])",
+            "/p:ProjectRepoOwner=dotnet",
+            "/p:ProjectRepoName=source-build",
+            "/p:ProjectRepoBranch=dev/release/2.1",
```
ryanbrandenburg: That's an interesting naming scheme, what does it signify?

dagood: At the moment the `dev/` prefix is a historical thing from the branching pattern for earlier releases. I believe we used it in the past to iterate on infrastructure while others did work with the existing infra to get a release out the door. Normally that would be `master`, but both branches used `release/...` submodules so that could have been misleading.

PR #274, opened 2018-03-26
Add release/2.1 trigger for AspNetCore

ryanbrandenburg:
Keep release/2.1 up to date in the same way that dev is.

CC @natemcmaster

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -691,6 +691,30 @@
         "vsoSourceBranch": "dev"
       }
     },
+    // Update dependencies in the aspnetcore repo's release/2.1 branch
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/product/cli/release/2.1/packages.semaphore"
+      ],
+      "action": "aspnet-dependency-update",
+      "delay": "00:05:00",
+      "actionArguments": {
+        "vsoSourceBranch": "master",
+        "vsoBuildParameters": {
+          "ScriptFileName": "scripts\\UpdateDependencies.ps1",
+          "Arguments": [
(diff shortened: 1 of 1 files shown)


natemcmaster:
cc @dagood @mmitche

Maestro/subscriptions.json
```diff
@@ -691,6 +691,30 @@
...
+        "vsoBuildParameters": {
+          "ScriptFileName": "scripts\\UpdateDependencies.ps1",
+          "Arguments": [
+            "-BuildXml",
+            "https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/product/cli/release/2.1/build.xml",
+            "-ConfigVars '",
+              "--GithubUpstreamBranch release/2.1",
+              "--GithubUsername dotnet-maestro-bot",
+              "--GithubEmail dotnet-maestro-bot@microsoft.com",
+              "--GithubToken=`$(`$Secrets[`'BotAccount-dotnet-maestro-bot-PAT`'])",
```
dagood: Why did you add `=`?

natemcmaster: I assume it was to workaround the goofy behavior of Microsoft.Extensions.Configuration.CommandLine.

ryanbrandenburg: When trying to test this script out locally by copying these arguments you won't have that value from $Secrets , do this will look like "--GithubToken --ArgumentWeAddOurselves" which messes up our parsing. By adding the = in that spot I prevent our argument parsing from going off the rails, either when we try to run it locally or if for some reason that Secret didn't exist. Is there any reason why it would be dangerous?

ryanbrandenburg: Nate has a much shorter explaination lol

PR #275, opened 2018-03-26
Fix syntax error in trigger params for aspnet-docker

natemcmaster:
cc @dagood


natemcmaster:
yes, I don't have write access

PR #278, opened 2018-03-28
Update aspnetcore from core-setup

ryanbrandenburg:
This should allow aspnetcore to take new dependencies when core-setup passes instead of waiting for final ProdCon. It depends on aspnet/Universe#1011.


ryanbrandenburg:
:up::date:. I've removed the release configs of the CoreSetup updater, but left the master ones in.

PR #286, opened 2018-04-09
Remove Core-Setup 2.0.0-2.0.7 & Update 2.0.0

wtgodbe:
The most recent Core-Setup 2.0.0 pipebuild was built with SourceBranch `release/2.0.0-2.0.7`, which caused a folder named `2.0.0-2.0.7` to be created inside of Core-Setup's build-info folder. This info should've gone into the `2.0.0` folder instead. This PR fixes that.

PTAL @weshaggard @dagood


wtgodbe:
I don't have write access to merge this, @dagood or @weshaggard can either of you do it (once approved)?

build-info/dotnet/core-setup/release/2.0.0/Last_Build_Packages.txt
```diff
@@ -1,45 +1,45 @@
-Microsoft.NETCore.App 2.0.8-servicing-26407-02
```
weshaggard: You are moving these versions backwards, do we really need to do that?

wtgodbe: I was trying to simulate what the auto-PR should have done - I'm fine with not touching the versions in these files if you think that's more correct.

dagood: Some downstream repo could have already upgraded to the new version and started depending on it, so downgrading is probably not the best.

wtgodbe:
Updated to address feedback

PR #316, opened 2018-06-18
Subscription added fron gitSyncCommitManager

Anipik:
Tested End to End with merge commit 
1 - change pushed to the shared path in the corefx repo

2 - Maestro collects the data about the commit and trigger the corresponding vsts build definition

3- This build definition stores the relevant Info in the azure table.

4- Mirroring tools then opens the PR for these changes from a fork account.

Opening through Fork
https://github.com/dotnet/coreclr/pull/18527
https://github.com/dotnet/coreclr/pull/18527

### DIFF
--- a/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Models/ModifiedFileModel.cs
+++ b/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Models/ModifiedFileModel.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Generic;
 using System.Diagnostics;
 
 namespace Microsoft.DotNet.Maestro.WebApi.Models
@@ -16,11 +17,14 @@ namespace Microsoft.DotNet.Maestro.WebApi.Models
 
         public string CommitId { get; set; }
 
+        public string CommitList { get; set; }
+
         public static bool TryParse(
             string repoFullName,
             string refSpec,
             string fullPath,
             Commit commit,
+            List<Commit> Commits,
             out ModifiedFileModel model)
         {
             // the file path goes like the following:
@@ -42,7 +46,8 @@ namespace Microsoft.DotNet.Maestro.WebApi.Models
                 RepoName = repoFullName,
                 BranchName = branchName,
                 FullPath = "https://github.com/" + string.Join("/", repoFullName, "blob", branchName, fullPath),
-                CommitId = commit.id
+                CommitId = commit.id,
+                CommitList = string.Join(";", Commits.ConvertAll( t => t.id ))
             };
 
             return true;
--- a/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Models/ModifiedFileModel.cs
+++ b/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Models/ModifiedFileModel.cs
@@ -4,6 +4,7 @@
 
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 
 namespace Microsoft.DotNet.Maestro.WebApi.Models
 {
@@ -24,7 +25,7 @@ namespace Microsoft.DotNet.Maestro.WebApi.Models
             string refSpec,
             string fullPath,
             Commit commit,
-            List<Commit> Commits,
+            IEnumerable<Commit> commits,
             out ModifiedFileModel model)
         {
             // the file path goes like the following:
@@ -47,7 +48,7 @@ namespace Microsoft.DotNet.Maestro.WebApi.Models
                 BranchName = branchName,
                 FullPath = "https://github.com/" + string.Join("/", repoFullName, "blob", branchName, fullPath),
                 CommitId = commit.id,
-                CommitList = string.Join(";", Commits.ConvertAll( t => t.id ))
+                CommitList = string.Join(";", commits.Select( c => c.id ))
             };
 
             return true;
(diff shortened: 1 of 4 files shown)


Anipik:
@eerhardt @danmosemsft @safern

Maestro/subscriptions.json
```diff
@@ -1567,6 +1573,25 @@
           "PB_SymbolExpirationInDays": "10950"
         }
       }
+    },
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/corefx/tree/master/src/Common/src/CoreLib/**/*",
+        "https://github.com/dotnet/corefx/tree/release/2.1/src/Common/src/CoreLib/**/*",
```
danmosemsft: You should probably make sure the tool works well against master branches before enabling your code changes against release branches. (Not sure whether that means the release branches can go here and filtering is done further down the chain)

Anipik: >  Not sure whether that means the release branches can go here and filtering is done further down the chain


Yes, the filtering is done in the mirroring tool. The  purpose of adding them here is too just have a record of commits that went to the shared path in the master and release branch. 
we can enable the tool for release branches later and it will pickup all the missing changes from this record

dagood:
Have you considered:

 1. What happens if a build fails? Will the commits be mirrored by subsequent builds? (Without special handling, my guess is "no".)
 2. What happens if GitHub or Maestro "misses" some commits, then starts picking up later pushes? Are the missed commits lost?
 3. What if two PRs are merged at about the same time, and the mirroring builds run in parallel and complete out of order?

Instead of `trigger-commitlist`, my suggestion would be to record the last mirrored commit in the repo receiving the mirroring commits. During the mirror build, you use `git` to check for all the commits that have happened to the relevant files between the last mirrored commit and the latest commit. This means you can easily tell where the mirror's at when looking at the target repo, and the considerations above don't matter.

Anipik:
>  my suggestion would be to record the last mirrored commit in the repo receiving the mirroring commits. 

We are currently using the lastSynchronizedCommit and it is leading to a lot of missing commits.

> What happens if GitHub or Maestro "misses" some commits, then starts picking up later pushes? Are the missed commits lost?

If maestro misses some commits or Webhooks fail, we will manually have to trigger the build definations for that commits. 

> What if two PRs are merged at about the same time, and the mirroring builds run in parallel and complete out of order?

The mirroring tool makes sure that there is only One Mirror PR per repo at one time.


@safern  any idea on how to tackle if there is some build failure or misses by the GitHub api or maestro ?

dagood:
> We are currently using the lastSynchronizedCommit and it is leading to a lot of missing commits.

Why is the best fix for this problem to move to a completely new way of doing things? Last synced commit seems purely more reliable to me.

safern:
> Why is the best fix for this problem to move to a completely new way of doing things? Last synced commit seems purely more reliable to me.

I think adding the commits that need to be mirrored into a database table and marking them as mirrored or not is a better approach than just having a lastSynchronizedCommit, we could add a mirrored date or something like that to the table and that way we could calculate the last commit we mirrored it; but having this new storage seems like a more robust way to do so, as we will have history and we will know what commits need to be mirrored and haven't.

dagood:
Having the mirroring information in a table sounds good to me, but it still seems like this new Maestro feature adds an unnecessary dependency on GitHub + Maestro being reliable. For this step:

> 3- This build definition stores the relevant Info in the azure table.

the build def can figure out the commits that are new since the last update to the table, just based on the latest commit.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Models/ModifiedFileModel.cs
```diff
@@ -42,7 +46,8 @@ public class ModifiedFileModel
                 RepoName = repoFullName,
                 BranchName = branchName,
                 FullPath = "https://github.com/" + string.Join("/", repoFullName, "blob", branchName, fullPath),
-                CommitId = commit.id
+                CommitId = commit.id,
+                CommitList = string.Join(";", Commits.ConvertAll( t => t.id ))
```
eerhardt: I would have expected this to be `Commits.Select(c => c.id)` .

Differences:

1. Select instead of ConvertAll.  (Select works on more collection types, not just List)
2. the param `c` instead of `t`.  I'm not sure what `t` stands for.

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Models/ModifiedFileModel.cs
```diff
@@ -16,11 +17,14 @@ public class ModifiedFileModel
...
         public string CommitId { get; set; }
 
+        public string CommitList { get; set; }
+
         public static bool TryParse(
             string repoFullName,
             string refSpec,
             string fullPath,
             Commit commit,
+            List<Commit> Commits,
```
dagood: In addition to @eerhardt's changes I'd also suggest changing this to `IEnumerable<Commit>`.

Anipik:
> the build def can figure out the commits that are new since the last update to the table, just based on the latest commit. 

@dagood 
Currently we use ```LibGit2Sharp``` to get the commits after the lastSynchronizedCommit But it will require for this build definition to clone the repos(corefx, coreclr, corert).
We can use the GitHub Rest api to get the commits after a certain date. But it will be similar to what maestro does with the webhook payload.

my suggestion is to store the LastSynchronizedCommit in the mirroring tool and  run a periodic sanity check which will insert any missing commits to the azure table and update the lastSynchronized commit.

dagood:
I see, I didn't realize you were doing this without having a repo clone. (I assume that means not even step 4 has a clone, so you never have full history access throughout the process.) Yeah, using the GitHub API to gather the info doesn't seem ideal. Doing something more complex and patching it up sounds more reasonable to me now. Thanks!

dagood:
Filed https://github.com/dotnet/core-eng/issues/3712 to track deploying the feature. I think I'll be able to help with that later today. (If we merge this PR now, you'll get `<trigger-commitlist>` passed literally to your build def until the deployment.)

dagood:
Seems like the perfect place for the check then, yeah, instead of just periodically. (Since it costs a negligible amount of time if you have the repo cloned with history.)

dagood:
This is deployed. Go ahead and file a core-eng FR issue and tag me if something doesn't seem to be working right.

Anipik:
Thanks :)

PR #320, opened 2018-06-25
Align 2.2 branches between aspnet and core-setup

natemcmaster:
The latest maestro PR to aspnet attempted to move aspnet ahead from 2.2 to 3.0. On aspnet, our dev branches are still appointed for 2.2 work (for now), so I'm updating the maestro trigger to only run for 2.2 changes.

@weshaggard do you have plans to enable publishing for the release/2.2 branches? I don't actually see anything for 2.2 under https://github.com/dotnet/versions/tree/master/build-info/dotnet/core-setup/release.

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -830,7 +830,7 @@
     // Update dependencies in the aspnetcore repo's dev branch after core-setup passes
     {
       "triggerPaths": [
-        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/core-setup/master/Latest.txt"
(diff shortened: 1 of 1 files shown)


natemcmaster:
FYI @ryanbrandenburg

natemcmaster:
Ok. Assuming that will happen at some point, is https://github.com/dotnet/versions/blob/master/build-info/dotnet/core-setup/release/2.2/Latest.txt going to be the right file to use as a trigger for 2.2 updates?

weshaggard:
Yes I would expect that to be the right file.

natemcmaster:
Thanks Wes and Davis!

PR #342, opened 2018-08-06
Updating WCF subscriptions.

StephenBonikowsky:
* We prefer not to update our CoreFx dependency in servicing branches unless specifically desired or required.
* Our Master branch is comparable to the CoreFx release/2.2 branch for now for purposes of dependencies and auto-updates.

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -1597,8 +1597,6 @@
     {
       "triggerPaths": [
         "https://github.com/dotnet/versions/blob/master/build-info/dotnet/standard/release/2.0.0/Latest.txt",
-        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/release/2.0.0/Latest.txt",
-        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/core-setup/release/2.0.0/Latest.txt",
         "https://github.com/dotnet/versions/blob/master/build-info/dotnet/wcf/release/2.0.0/Latest.txt"
       ],
       "action": "wcf-general",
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
@dagood could you add reviewers as needed please? Thanks.

dagood:
I'm enough to do the review + merge. Looks like you need to rebase first though.

Maestro/subscriptions.json
```diff
@@ -942,8 +942,6 @@
     {
       "triggerPaths": [
         "https://github.com/dotnet/versions/blob/master/build-info/dotnet/standard/release/2.0.0/Latest.txt",
-        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/release/2.0.0/Latest.txt",
-        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/core-setup/release/2.0.0/Latest.txt",
```
dagood: Keep in mind these only affect the frequency of updates (by triggering them), and the update PR contents depend only on dependencies.props. E.g. `RemoteDependencyBuildInfo Include="CoreFx"` and `XmlUpdateStep Include="CoreFx"` and the Core-Setup ones should be disabled/commented/removed from https://github.com/dotnet/wcf/blob/release/2.0.0/dependencies.props to match this trigger change.

The same goes for the others--like in WCF `master`, you'll have to update dependencies.props to point to `release/2.1` build-infos.

dagood: I think I'd actually suggest doing the WCF PRs first, since that describes what will happen during upgrades, then coming back to make the triggers match.

StephenBonikowsky: Yup, I'm going through the dependencies.props files in each of our branches to do what we want. Since I had to update at least the trigger paths for our Master branch I thought I'd "cleanup" all the entries for WCF.

PR #343, opened 2018-08-06
Fix invalid trigger path for WCF Master subscription.

StephenBonikowsky:
Made a mistake in my previous PR with the trigger path.


StephenBonikowsky:
@dagood

PR #358, opened 2018-08-30
Transition VSTS account name to dnceng

chcosta:
@mmitche @dagood PTAL

### DIFF
--- a/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Web.config
+++ b/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Web.config
@@ -8,7 +8,7 @@
     <add key="WebhookSecretToken" value="" />
     <add key="KeyVaultClientId" value="" />
     <add key="KeyVaultSecret" value="" />
-    <add key="dotnet.visualstudio.com-KeyVaultSecretName" value="" />
+    <add key="dnceng.visualstudio.com-KeyVaultSecretName" value="" />
     <add key="devdiv.visualstudio.com-KeyVaultSecretName" value="" />
     <add key="VstsUsername" value="" />
     <add key="SubscriptionsUrl" value="https://raw.githubusercontent.com/dotnet/versions/master/Maestro/subscriptions.json" />
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -24,9 +24,9 @@
       "vsoProject": "DevDiv",
       "buildDefinitionId": 8289
     },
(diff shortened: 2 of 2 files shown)


chcosta:
I don't have merge permissions for this repo, can someone please merge as soon as CI passes?

Maestro/src/Microsoft.DotNet.Maestro.WebApi/Web.config
```diff
@@ -8,7 +8,7 @@
     <add key="WebhookSecretToken" value="" />
     <add key="KeyVaultClientId" value="" />
     <add key="KeyVaultSecret" value="" />
-    <add key="dotnet.visualstudio.com-KeyVaultSecretName" value="" />
+    <add key="dnceng.visualstudio.com-KeyVaultSecretName" value="" />
```
dagood: This won't take effect until deployment (which is manual), so I think you (or FR, or someone) will need to deploy and update the value in the app service. https://github.com/dotnet/core-eng/blob/master/Documentation/Maestro/deploy.md

dagood: Not sure. I'd expect an auth error when it tries to trigger the job (or maybe it would just hit the 404).

dagood:
Actually nevermind about merging--I can't say for sure that this won't break the non-VSTS scenarios before the deployment. My concern is that it would try to use the setting and get an exception during setup, maybe making all other triggers stop working.

@mmitche, since you set up the vault PAT stuff, maybe you know better.

chcosta:
I will start working on the deploy at 11:30 am PST.  @dagood, @mmitche , let me know if there is any concern with that.

dagood:
I looked a bit at the auth flow and I'm more confident now it's safe to merge this without interrupting devdiv instance triggers. The credential-loading `Config.Instance.GetPassword` only happens all the way in the VSTS queue method:

https://github.com/dotnet/versions/blob/82ad5fa276bcec84548ba7e31d747be0164f99f5/Maestro/src/Microsoft.DotNet.Maestro/Services/VsoService.cs#L40-L41

I'd expect devdiv triggers to be fine because those lookups wouldn't be affected. Individual dotnet/dnceng instance VSTS triggers should fail until the deploy and update happens, but they would fail anyway.

dagood:
Confirmed: devdiv triggers still working, dnceng triggers getting `Could not locate a secret name for dnceng.visualstudio.com VSTS instance (looked up config key dnceng.visualstudio.com-KeyVaultSecretName)` which is expected until deployment/app service change.

PR #370, opened 2018-10-02
Remove IBC update line

DrewScoggins:
Since we are not currently updating the IBC package and there is nothing
in the location this is causing a crash in the maestro job. I am going
to remove this until we have the data pushing then we can add this back.


DrewScoggins:
PTAL @mmitche  @dagood

PR #371, opened 2018-10-09
Add tracking for IBC package

DrewScoggins:
I removed this as we were not producing IBC pacakges at the time, but
now that we have IBC data being regularly updated we should switch this
trigger back on.


DrewScoggins:
@mmitche @dagood @adiaaida PTAL

PR #408, opened 2018-12-04
Manually update projectn-tfs-testilc/master

zamont:
Manually update projectn-tfs-testilc/master to keep it in sync with projectn-tfs/master that was manually updated to version beta-27301-01.


zamont:
@dagood @wtgodbe this fixes the issues with the auto-update in https://github.com/dotnet/corefx/pull/33783

PR #409, opened 2018-12-06
Manually correct the list of latest corefx packages on master

natemcmaster:
These packages are apparently no longer produced by the master branch. This messes up our automation everytime we try to update.

cc @sebastianros


natemcmaster:
Oops, my bad.

SebastianRos:
Nah, it's okay. I'm getting used to it

SebastianRos:
Not many. I think I was accidentally mentioned about two or three times since I signed up in 2015.

PR #415, opened 2019-01-18
Add dotnet/test-templates repo.

singhsarab:
Adding the dotnet/test-templates repo to the list of repos in subscriptions.json


singhsarab:
@tmat Please review the change.

tmat:
you might want to add `release` as well, in case we need to create release branches

PR #416, opened 2019-01-23
Adding subscription of WCF branch for dnceng internal mirror.

StephenBonikowsky:
* During the adoption phase will be doing all the work out of the AdoptArcade branch before moving it into master.

Auto format removed the comma from line 835/836, I believe this is how it should be anyway, correct?


StephenBonikowsky:
@chcosta @jonfortescue 
I can't add reviewers, could you assist with this?
Thanks.

PR #425, opened 2019-02-13
Manually publish BuildTools preview1-03713-01

tannergooding:
CC. @dagood. The publishing step is still broken for the build-tools job, manually publishing it.


tannergooding:
Corresponding job is https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2410427

PR #432, opened 2019-02-19
Added trigger info for the aspnet/Razor.VSCode repo

mkArtakMSFT:
This step is required as part of the work listed https://github.com/aspnet/AspNetCore-Internal/issues/1577

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -826,6 +826,7 @@
         "https://github.com/aspnet/EntityFrameworkCore/blob/release/**/*",
         "https://github.com/aspnet/Extensions/blob/master/**/*",
         "https://github.com/aspnet/Extensions/blob/release/**/*",
+        "https://github.com/aspnet/razor.vscode/blob/master/**/*",
         "https://github.com/aspnet/Scaffolding/blob/master/**/*",
         "https://github.com/aspnet/Scaffolding/blob/release/**/*",
         "https://github.com/aspnet/websdk/blob/master/**/*",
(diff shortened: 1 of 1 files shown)


mkArtakMSFT:
@natemcmaster can you please confirm I got this right?

Maestro/subscriptions.json
```diff
@@ -826,6 +826,7 @@
         "https://github.com/aspnet/EntityFrameworkCore/blob/release/**/*",
         "https://github.com/aspnet/Extensions/blob/master/**/*",
         "https://github.com/aspnet/Extensions/blob/release/**/*",
+        "https://github.com/aspnet/razor.vscode/blob/master/**/*",
```
natemcmaster: I think this will work, but just to be safe, would be good to match the casing of the repo.

image

dagood: It will, but agree for cleanliness:

https://github.com/dotnet/versions/blob/82ad5fa276bcec84548ba7e31d747be0164f99f5/Maestro/src/Microsoft.DotNet.Maestro.WebApi/Models/SubscriptionsModel.cs#L83-L92

Could consider including the release pattern trigger. Might as well, if there are ever going to be release branches building in the repo. (There doesn't need to be something already there.)

natemcmaster: added the release/* pattern.

mkArtakMSFT:
Thanks for looking into this, guys.
I just discovered that we're going to merge the aspnet/Razor.VSCode into the aspnet/aspnet-tooling repo and that eliminates the need for this change.

PR #439, opened 2019-02-28
Remove PGO, IBC, CoreFX, and Core-Setup flow to CoreCLR

hoyosjs:
Remove PGO, IBC, CoreFX, and Core-Setup flow to CoreCLR as the now flow using Darc/Maestro++

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -166,11 +166,8 @@
     // Update dependencies in CoreCLR master
     {
       "triggerPaths": [
-        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/Latest.txt",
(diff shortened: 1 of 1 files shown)


dagood:
PGO and IBC are still configured to auto-update using VersionTools, this dotnet/versions PR will only make the updates trigger less often. https://github.com/dotnet/coreclr/blob/062597365c5fe0cbab66aa11447e8e6459179cd6/dependencies.props#L114-L123

It's not very important to remove triggers, the only effect of excessive triggers is that a dotnet-maestro-bot auto-update PR might be regenerated when it didn't really need to be. The config in the repo controls what's upgraded and when.

hoyosjs:
Oh ok, see dotnet/coreclr#22915. We just keep getting the PR of optimization bundled with CoreClr updates, which was inconvenient because they got triggered enough to trip CI before completing. Once the mentioned PR goes in all should be good then. Thanks!

PR #442, opened 2019-02-28
Add auto-merge from Master into Release/3.0

wtgodbe:
Please don't merge this until the release/3.0 branches are open for Preview4 (I'll poke at this PR once that's the case)

Adds auto-merges from master into release/3.0 for CoreClr, CoreFx, and Core-Setup.

@mmitche @dagood PTAL

CC @danmosemsft

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -1857,7 +1857,7 @@
       "triggerPaths": [
         "https://github.com/dotnet/coreclr/blob/master/**/*",
         "https://github.com/dotnet/corefx/blob/master/**/*",
-        "https://github.com/dotnet/core-setup/blob/master**/*",
+        "https://github.com/dotnet/core-setup/blob/master/**/*",
       ],
       "action": "dotnet-arcade-general",
       "actionArguments": {
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -1852,6 +1852,31 @@
         }
       }
     },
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -1852,6 +1852,31 @@
         }
       }
     },
+    // Automate opening PRs to merge dotnet org repos' master changes into release/3.0 branches
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/coreclr/blob/master/**/*",
+        "https://github.com/dotnet/corefx/blob/master/**/*",
+        "https://github.com/dotnet/core-setup/blob/master**/*",
```
dagood:
```suggestion
        "https://github.com/dotnet/core-setup/blob/master/**/*",
```

wtgodbe:
For 2.1/2.2 I would generally resolve those myself manually, I don't mind doing the same here. Unless there's a way to exclude specific files (Versions.props, version.details.xml, dependencies.props)?

dagood:
Thanks, I do think it's useful even if it's always conflicting--it's at least a reminder that shows you what's missing at that point in time. 😄 

I can only imagine something disruptive like adding a duplicate of the version files to `master` and updating/using those instead. Then, since the originals are left alone, there wouldn't be any conflicts with the old copy that `release/3.0` uses. Not very palatable.

Knowing Git, there's probably some built in trick, I haven't looked for one though.

PR #445, opened 2019-03-11
Update Core* repos from buildtools 3.0 rather than master

wtgodbe:
I've just branched buildtools to release/3.0 (as per discussions on Friday), so we should update Core* dependencies from there rather than from master. https://github.com/dotnet/buildtools/commit/2912fd36eaffe4e998073044d0be9319ff5213dd

CC @danmosemsft @leecow @mmitche @dagood

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -421,7 +421,7 @@
     // Update dependencies in CoreCLR release/3.0
     {
       "triggerPaths": [
-        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/buildtools/master.txt",
+        "https://github.com/dotnet/versions/blob/master/build-info/dotnet/buildtools/release/3.0/Latest.txt",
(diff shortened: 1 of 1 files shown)


wtgodbe:
@dagood I might be misunderstanding something here, is there an additional change needed in the Core* repos to complete this task?

dagood:
Yeah, each one needs a change to `dependencies.props` to pull the right build-info. For example, CoreFX:

https://github.com/dotnet/corefx/blob/e433981ea570928efbb1587cc6735c90c1881a58/eng/dependencies.props#L82-L85
```xml
    <RemoteDependencyBuildInfo Include="BuildTools">
      <BuildInfoPath>$(BaseDotNetBuildInfo)buildtools/$(DependencyBranch)</BuildInfoPath>
      <CurrentRef>$(BuildToolsCurrentRef)</CurrentRef>
    </RemoteDependencyBuildInfo>
```

Either `DependencyBranch` needs to be `release/3.0` (may affect other dependencies) or the buildtools dependency in particular needs to be changed to `release/3.0`.

(This is because Maestro and VersionTools are extremely decoupled.)

PR #455, opened 2019-04-02
Add aspnet/EntityFramework6

bricelam:


Maestro/subscriptions.json
```diff
@@ -1250,6 +1252,7 @@
       "triggerPaths": [
         "https://github.com/aspnet/AspNetCore/blob/release/3.0-preview3/**/*",
         "https://github.com/aspnet/AspNetCore-Tooling/blob/release/3.0-preview3/**/*",
+        "https://github.com/aspnet/EntityFramework6/blob/release/3.0-preview3/**/*",
```
bricelam: Adding now so it doesn't get missed in preview4

PR #457, opened 2019-04-03
Mirror aspnet/AspNetCore-Internal

anurse:
We use this repo for some infrastructure tools we currently run through TeamCity and want to start using AzDO for this. We need to mirror to dnceng/internal so that we can use it there :)

cc @Eilon @jkotalik @dougbu @ryanbrandenburg

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -823,6 +823,7 @@
         "https://github.com/aspnet/AspLabs/blob/release/**/*",
         "https://github.com/aspnet/AspNetCore/blob/master/**/*",
         "https://github.com/aspnet/AspNetCore/blob/release/**/*",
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -823,6 +823,7 @@
         "https://github.com/aspnet/AspLabs/blob/release/**/*",
         "https://github.com/aspnet/AspNetCore/blob/master/**/*",
         "https://github.com/aspnet/AspNetCore/blob/release/**/*",
+        "https://github.com/aspnet/AspNetCore-Internal/blob/master/**/*",
```
dagood: If there's any chance of release branches being useful, a proactive trigger here wouldn't break anything.

PR #479, opened 2019-06-12
Add subscription for aspnet/HttpRepl

tlmii:
aspnet/HttpRepl is being split off from aspnet/AspLabs


tlmii:
@mmitche Can you review?

PR #480, opened 2019-06-13
Remove subscription for dotnet-docker/nightly

mthalman:
This is being replaced by our own system for managing base images.  See https://github.com/dotnet/docker-tools/issues/59


mthalman:
CC @dagood for review

PR #498, opened 2019-08-23
Remove base image pipeline trigger for dotnet-docker

mthalman:
Remove the automatic build trigger for the dotnet-docker repo when base image changes occur.

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -1235,33 +1235,6 @@
         }
       }
     },
-    // Trigger official build of the dotnet-docker repo's master branch when a base image is changed
-    {
-      "triggerPaths": [
-        "https://github.com/dotnet/versions/blob/master/build-info/docker/alpine/3.7.txt",
-        "https://github.com/dotnet/versions/blob/master/build-info/docker/alpine/3.8.txt",
-        "https://github.com/dotnet/versions/blob/master/build-info/docker/alpine/3.9.txt",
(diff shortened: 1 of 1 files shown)


Maestro/subscriptions.json
```diff
@@ -1235,33 +1235,6 @@
         }
       }
     },
-    // Trigger official build of the dotnet-docker repo's master branch when a base image is changed
```
MichaelSimons: It looks like the ["action" declarations for both dotnet-docker nightly and master](https://github.com/dotnet/versions/blob/2dae12e716b82bc6509d5459c3dc38f728ab2252/Maestro/subscriptions.json#L87) can be removed as well.  Can you please clean them up with this PR?

mthalman: Nightly is still hooked up to 2.x changes even.  Do we still want that configured?

MichaelSimons: That is a different action though isn't it?  https://github.com/dotnet/versions/blob/2dae12e716b82bc6509d5459c3dc38f728ab2252/Maestro/subscriptions.json#L81

mthalman: That's the action for `dotnet-docker-update-dependencies` which we still want around to track digest updates for reference.  There's also the action for `dotnet-docker-nightly` which is hooked up to 2.x changes.  So the only one we can get rid of is `dotnet-docker`.

MichaelSimons: Discussed offline and resolved - the `dotnet-docker` and `dotnet-docker-nightly` can be cleaned up.  `dotnet-docker-update-dependencies` will be preserved.

PR #503, opened 2019-09-05
Add rules for WPF branches

vatsan-madhavan:
Adding rules for automatically porting WPF changes from...

- ... release/3.0 -> release/3.1
- ... release/3.1 -> master 

/cc @mmitche, @rladuca

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -1805,7 +1805,8 @@
         "https://github.com/dotnet/coreclr/blob/release/3.0/**/*",
         "https://github.com/dotnet/corefx/blob/release/3.0/**/*",
         "https://github.com/dotnet/core-setup/blob/release/3.0/**/*",
-        "https://github.com/dotnet/source-build/blob/release/3.0/**/*"
+        "https://github.com/dotnet/source-build/blob/release/3.0/**/*",
+        "https://github.com/dotnet/wpf/blob/release/3.0/**/*"
       ],
       "action": "github-dnceng-branch-merge-pr-generator",
       "actionArguments": {
@@ -1815,6 +1816,22 @@
           "GithubRepoName": "<trigger-repo>",
           "HeadBranch": "<trigger-branch>",
           "BaseBranch": "release/3.1"
+        }
+      }
+    },
+    // Automate opening PRs to merge dotnet org repos' release/3.1 changes into master branches
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/wpf/blob/release/3.1/**/*"
+      ],
+      "action": "github-dnceng-branch-merge-pr-generator",
+      "actionArguments": {
+        "vsoSourceBranch": "master",
+        "vsoBuildParameters": {
+          "GithubRepoOwner": "dotnet",
+          "GithubRepoName": "<trigger-repo>",
+          "HeadBranch": "<trigger-branch>",
+          "BaseBranch": "master"
         }
       }
     },


Maestro/subscriptions.json
```diff
@@ -1818,6 +1819,22 @@
...
+        "https://github.com/dotnet/wpf/blob/release/3.1/**/*"
+      ],
+      "action": "github-dnceng-branch-merge-pr-generator",
+      "actionArguments": {
+        "vsoSourceBranch": "master",
+        "vsoBuildParameters": {
+          "GithubRepoOwner": "dotnet",
+          "GithubRepoName": "<trigger-repo>",
+          "HeadBranch": "<trigger-branch>",
+          "BaseBranch": "master"
```
rladuca: What is this going to do for DARC flow into release/3.1?  Will it try to merge those version updates into master?

vatsan-madhavan: Yes, it will. There is no exclusion for eng\common, eng\Version*, global.json etc. 

We'll have to decide how best to curate PR's initiated by maestro-bot. We can either have the convenience of everything being auto-merged (i.e, no reviews required), or we can review everything and reject some of these updates (i.e., change policy to reviews being required).

vatsan-madhavan: @mmitche, can the `triggerPaths` here be rewritten to be more specific, like this:

```json
    "https://github.com/dotnet/wpf/blob/release/3.0/src/**/*",
    "https://github.com/dotnet/wpf/blob/release/3.0/packaging/**/*",
    "https://github.com/dotnet/wpf/blob/release/3.0/Documentation/**/*",
    "https://github.com/dotnet/wpf/blob/release/3.0/eng/*",
    "https://github.com/dotnet/wpf/blob/release/3.0/eng/WpfArcadeSdk/**/*",
    "https://github.com/dotnet/wpf/blob/release/3.0/*.sln",
    "https://github.com/dotnet/wpf/blob/release/3.0/*.props",
    "https://github.com/dotnet/wpf/blob/release/3.0/*.targets",
    "https://github.com/dotnet/wpf/blob/release/3.0/*.config",
    "https://github.com/dotnet/wpf/blob/release/3.0/*.yml",
    "https://github.com/dotnet/wpf/blob/release/3.0/*.cmd",
    "https://github.com/dotnet/wpf/blob/release/3.0/*.ps1",
    "https://github.com/dotnet/wpf/blob/release/3.0/*.sh",
    "https://github.com/dotnet/wpf/blob/release/3.0/*.md",
    "https://github.com/dotnet/wpf/blob/release/3.0/*.txt"
```

rladuca: If we move this to manual PRs, we might have to shut this down during Open Sourcing.  If we do the same sort of "let flow break then merge updates into DARC PRs" thing we usually do, won't we potentially get a flow that contains both DARC and OSS changes?  I suppose that's a one time cost where we can just back out the DARC changes, but we'd need to make sure we remember the way we choose to handle it.

mmitche: @dagood May know better, but I don't think so.  There will be merge conficts when it tries to merge.

Anyways, we want this behavior.  We want to be able to say that all commits from 3.0 have been merged into 3.1 and then into master, even if the process of merging those commits involves resolving a bunch of conflicts.  The ability to say that a sha with a fix has been integrated into all branches is important. If you only merge certain paths, you're not merging the commit, you're just making new changes.

The behavior could be smarter.  For example, an auto-merging policy that preferences the content of certain files from the target branch.

dagood: Adding subpaths works, but `*` is not supported. And it's only a trigger, it doesn't change what the underlying AzDO build actually does, so the auto merge PRs would end up the same anyway. We also can't have the underlying build filter out the files because we do need an actual merge commit, as Matt's saying.

Something that came to mind a while back is that we could give each branch a new set of version files, so that versions from lesser branches merge cleanly into the newer ones. Then, the lower versions are ignored during the build. For example, 3.0 gets a dependency update, that change goes into, say, `eng/release/3.0/Versions.props`, and the resulting commit gets merged into the `release/3.1` branch. But when `release/3.1` builds, it uses `eng/release/3.1/Versions.props`. All the way up: a branch with N branches behind it has N copies of the Versions.props file. Just a fun idea though, too much noise to swallow. :smile:

Having the branch merge script add a smart merge commit on top of its branch before submitting the PR would be more reasonable.

Maestro/subscriptions.json
```diff
@@ -1818,6 +1819,22 @@
         }
       }
     },
+    // Automate opening PRs to merge dotnet org repos' release/3.1 changes into master branches
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/wpf/blob/release/3.1/**/*"
```
dagood: This suggests you're going with the approach of merging any change into the oldest applicable branch first, then letting it flow to `master` eventually. In the Core repos changes are instead merged into `master` first, then ported into the oldest applicable branch. Repos do it either way--as far as I'm aware it's a matter of opinion. Wondering if it's a conscious choice for WPF in particular.

vatsan-madhavan: We would want commits to be merged into simultaneously into `master `as well as `release/* `branches as a best-practice. This is a back-stop to guarantee "forward porting" of fixes. For e.g., suppose we take a small point-fix into `release/3.0` directly that for some reason was missed being committed to `master `directly, this rule would ensure that it would flow into `release/3.1` and also `master`. OTOH, if a fix was made into `master`, then cherry-picked into `release/3.0`, then it would simply flow into `release/3.1`(useful), and and ensuing merge from `release/3.1` -> `master `would be empty (noise - but no harm done). 

Since the churn in `release/*` branches is generally low, it seems like a small price to pay (dealing with occasional empty merge PR's from `release/3.1 -> master`) to guarantee that we don't forget to forward-port changes inadvertently.

dagood: Seems like a reasonable approach, this can cause conflicts when some commit is ported from `master` to `3.1` after the fact, if some other commit has changed the same line. Someone has to go figure out the conflict. Ran an example to prove it to myself, in case the scenario isn't clear:

image

`1` and `2` just change the only line in a file to that number. The conflict is:

```
<<<<<<< HEAD
2
=======
1
>>>>>>> 3.1
```

Not sure if this is something you haven't considered. Being able to tell if a change is properly everywhere based on whether it's contained in certain Git trees would certainly be useful.

vatsan-madhavan: thx - this is a good caveat to remember. I still think I'd like to onboard it this way. If it turns out to be a hassle, we can always change/weak this process, right? if that's ok, then lets go ahead and merge this.

PR #517, opened 2019-10-03
Delete obsolete image info file

mthalman:
Deleting since this is file is no longer used.


mthalman:
> LGTM - I assume you will wait to merge until all Docker repos have been updated to use latest eng/common?

Yep, they've all been updated.  Merging this change now.

PR #525, opened 2019-11-01
Disable 3.0 -> 3.1 auto-merges during Preview3 build window

wtgodbe:
CC @Anipik @livarcocc @nguerrera @dagood @dougbu 

I'll merge this Sunday, then revert it after we declare a final build for GA

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -1193,26 +1193,6 @@
         }
       }
     },
-    // Automate opening PRs to merge cli 3.0.1xx to 3.1.1xx
-    {
-      "triggerPaths": [
-        "https://github.com/dotnet/cli/blob/release/3.0.1xx/**/*",
-        "https://github.com/dotnet/toolset/blob/release/3.0.1xx/**/*",
-        "https://github.com/dotnet/sdk/blob/release/3.0.1xx/**/*",
-        "https://github.com/dotnet/core-sdk/blob/release/3.0.1xx/**/*"
-      ],
-      "action": "github-dnceng-branch-merge-pr-generator",
-      "actionArguments": {
-        "vsoSourceBranch": "master",
-        "vsoBuildParameters": {
-          "GithubRepoOwner": "dotnet",
-          "GithubRepoName": "<trigger-repo>",
-          "HeadBranch": "<trigger-branch>",
-          "BaseBranch": "release/3.1.1xx",
-          "ExtraSwitches": "-QuietComments"
-        }
-      }
-    },
     // Automate opening PRs to merge cli 3.1.1xx to master
     {
       "triggerPaths": [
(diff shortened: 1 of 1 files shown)


nguerrera:
Do you also want to do this for cli/sdk/toolset/core-sdk? We just moved to this bot but our entries are elsewhere since our branches are differently named.

wtgodbe:
> Do you also want to do this for cli/sdk/toolset/core-sdk? We just moved to this bot but our entries are elsewhere since our branches are differently named.

Yes, thanks for the heads up. Just pushed a commit

nguerrera:
Seems that we should have an enabled: false or something instead of making revert noise in the history

wtgodbe:
> Seems that we should have an enabled: false or something instead of making revert noise in the history

@dagood do you know if there's a way to do this?

nguerrera:
Makes it hard to reference what is authored at any time. 

Even better would be something on the backend that can disable.

dagood:
I think the best you can do with what's implemented right now is comment out the triggers or something along those lines. The code is at https://github.com/dotnet/versions/tree/master/Maestro/src if you'd like to browse/contribute. Note that core-eng has to manually deploy changes and they may have a high bar, although this change would be trivial in terms of code.

Core-eng is targeting getting this functionality off Maestro somehow. https://github.com/dotnet/arcade/issues/405 (don't know if there's an active tracker)

dagood:
> Even better would be something on the backend that can disable.

The build def or branch merge script could do some additional logic to no-op itself based on external data. Maybe search for the last auto-merge PR with the same config, see if it has a "don't" tag?

Could just add another json file to this repo that contains streamlined branch-merge-specific enable/disable info, something like that.

nguerrera:
Is this already off somehow, I was waiting for a merge from 3.0 -> 3.1 of core-sdk to get some goodness for our CI flakiness.

wtgodbe:
It shouldn't be, how recently was the commit merged into 3.0? If it was before you added the SDK repos here (3 days ago), the builds wouldn't have been triggered. They only run when a new commit gets merged into a trigger branch

nguerrera:
It was this morning. The triggers have been working.

wtgodbe:
Weird. You can always run https://dnceng.visualstudio.com/internal/_build?definitionId=350 manually

nguerrera:
Oh, it's failing.

nguerrera:
https://dnceng.visualstudio.com/internal/_build/results?buildId=411827&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a

nguerrera:
Oh, if I merge an existing PR that it failed to update, then it doesn't trigger again. for things that came up when it failed to update I guess. Hmm.

nguerrera:
OK, I triggered it, and I see the issue. I will log something, but I'm unblocked.

PR #526, opened 2019-11-03
[NO MERGE] Revert "Disable 3.0 -> 3.1 auto-merges during Preview3 build window"

wtgodbe:
Reverts dotnet/versions#525

Re-enable auto-merges from 3.0 into 3.1. Will merge this after GA.

CC @dagood @nguerrera @Anipik


nguerrera:
When can these come back?

wtgodbe:
> When can these come back?

That's a tricky one, as the overlap between 3.0.2 being open and 3.1.1 being open is pretty much nil. I do think there's value in firing these off, though. I think it'd be fine to re-enable it for now, allow PRs to be opened across the stack (and merged after we update branding), then disable again next week when we lock down 3.1.1.

nguerrera:
I mostly wanted to make sure there's a plan here and this wasn't simply forgotten. Exactly what I thought would happened, happened: I went to go reference the subscription to show someone something and saw it was gone. cc @wli3 

I still think there needs to be another way to turn it on off than deleting the code.

wtgodbe:
> I still think there needs to be another way to turn it on off than deleting the code.

I completely agree, it isn't sustainable doing it this way. I just bumped https://github.com/dotnet/arcade/issues/405, hopefully we'll get some traction in this

PR #534, opened 2019-11-22
Merge 3.1 -> blazor-wasm -> master in AspNet repos

wtgodbe:
@JunTaoLuo @dougbu PTAL

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -1218,9 +1218,6 @@
     // Automate opening PRs to merge aspnet release/3.1 branches back to master
     {
       "triggerPaths": [
-        "https://github.com/aspnet/AspNetCore/blob/release/3.1//**/*",
-        "https://github.com/aspnet/AspNetCore-Tooling/blob/release/3.1//**/*",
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
cc @javiercn maybe this is the secret sauce we need.

javiercn:
I think do

dougbu:
Note this will not
1. merge directly from our 'release/2.1' branches to 'release/3.1'
2. do anything about issues w/ merges from 'release/3.1' to 'blazor-wasm' where we have 'blazor-wasm' branches
  - I've seen such merges but the branches might be too far out of sync due to incorrect merges or some such

wtgodbe:
I've updated the PR to also merge directly from 2.1 -> 3.0, skipping 2.2. @javiercn @JunTaoLuo I also don't think this will fix issues merging 3.1 -> blazor-wasm. It will just make it so that we go from 3.1 -> blazor-wasm -> master, rather than 3.1 going to both places.

wtgodbe:
@javiercn @JunTaoLuo - this should now just merge rel/3.1 -> blazor-wasm -> master for AspNetCore.

PR #539, opened 2019-11-26
Disable 3.0 -> 3.1 automerges during 3.0.2/3.1.1

wtgodbe:
CC @mmitche @nguerrera


wtgodbe:
@dagood is this the right way to comment these out?

dagood:
This looks like it'll work fine, I'll leave it up to you whether this seems more maintainable vs. removing and reverting later.

Seems reasonable to me to go this way: comments might make it easier to toggle these without merge conflicts. The schema doesn't really change, so it seems unlikely to go out of date (I think a typical concern with commenting out code).

PR #543, opened 2019-12-16
Add llvm-project to list of repos mirrored to internal

directhex:


directhex:
We don't do any integration work against LLVM master. There's the supported branch (release/6.x) and the WIP branch (release/9.x last time I checked)

PR #571, opened 2020-03-23
Convert image info for nightly branch to new schema

mthalman:
Update the image info file for the dotnet-docker/nightly branch to use the new schema defined by https://github.com/dotnet/docker-tools/issues/438.


mthalman:
Hold off on review until after today's release since the updates will need to be merged in and converted.

build-info/docker/image-info.dotnet-dotnet-docker-nightly.json
```diff
@@ -1,2575 +1,2172 @@
...
+          },
+          "platforms": [
+            {
+              "dockerfile": "2.1/aspnet/nanoserver-1809/amd64/Dockerfile",
+              "simpleTags": [
+                "2.1.16-nanoserver-1809",
+                "2.1-nanoserver-1809"
+              ],
+              "digest": "sha256:b23288ea3e92a12a584b8279b37dd5e7eb914dd29fd6a5d5bc7211b5248c9ecc",
+              "baseImageDigest": "mcr.microsoft.com/windows/nanoserver@sha256:5de6bd32bd453d60c8f549d28845552e89ad3652566e141ac82023b6ba10374d",
```
MichaelSimons: This is wrong.  I noticed it is wrong in the existing data as well.  It looks like the logic in the image-info-schema branch will correct this.  I don't think anything needs to be done here but now that info will include internal references, does the hello build kitty infra ignore "internal" base images?  By that I mean, it shouldn't retrieve the internal image digests from the registries. There is no need to monitor them.

build-info/docker/image-info.dotnet-dotnet-docker-nightly.json
```diff
@@ -1,2575 +1,2172 @@
...
+              "architecture": "arm32",
+              "created": "2020-03-23T13:30:24.5166081Z"
+            },
+            {
+              "dockerfile": "5.0/aspnet/buster-slim/arm64v8/Dockerfile",
+              "simpleTags": [
+                "5.0-buster-slim-arm64v8",
+                "5.0.0-preview-buster-slim-arm64v8"
+              ],
+              "digest": "sha256:7ef29c87cdd67a7b13388f12d85d5b047fe03644783cf14cc603f791e0133ecf",
```
MichaelSimons: Is it inconsistent that the image digest is just the digest where as the baseImageDigest is repo qualified?  The image is scoped to the containing repo which defines the scope.  I think the answer is does it help the consumer to qualify the image digest?

build-info/docker/image-info.dotnet-dotnet-docker-nightly.json
```diff
@@ -1,2575 +1,2172 @@
...
+              "dockerfile": "5.0/aspnet/nanoserver-1809/amd64/Dockerfile",
+              "simpleTags": [
+                "5.0.0-preview-nanoserver-1809",
+                "5.0-nanoserver-1809"
+              ],
+              "digest": "sha256:903eef3902d89c081d05b9dbbd14577836063b9833c350ade340305b8a29b390",
+              "osType": "Windows",
+              "osVersion": "Windows Server 2019",
+              "architecture": "amd64",
+              "created": "2020-03-23T13:24:44.6081702Z"
```
MichaelSimons: Is the conversion tool defaulting this time?  If so that feels wrong.  I think the created time should be left empty.

build-info/docker/image-info.dotnet-dotnet-docker-nightly.json
mthalman: What is wrong about this?  This isn't an in internal base image; it's the base Windows image.

build-info/docker/image-info.dotnet-dotnet-docker-nightly.json
mthalman: That's a fair question.  From a human readability standpoint, I can see how it would be useful to have it fully qualified because a lot of scrolling could be required to find the repo name.

build-info/docker/image-info.dotnet-dotnet-docker-nightly.json
mthalman: No, it's pulling the image and getting the created time from that.

build-info/docker/image-info.dotnet-dotnet-docker-nightly.json
MichaelSimons: My mistake, I got mixed up thinking about `installer` stage.  I think my question about hello build kitty applies.  It looks like now we always include the base image where as before we only included external base images.

mthalman: The `BaseImageDigest` value for a base image that is internally built is not currently stored.  This is because the base image digest is retrieved before the images have been pushed, so it ends up being null.  I've logged https://github.com/dotnet/docker-tools/issues/447 to track this.

build-info/docker/image-info.dotnet-dotnet-docker-nightly.json
mthalman: I've logged https://github.com/dotnet/docker-tools/issues/448 to track this.

mthalman:
@MichaelSimons - This is ready for review again.

PR #613, opened 2020-06-22
Removing stale SDK interbranch merge rules

sfoslund:


sfoslund:
@wli3 I'm removing this since it's outdated, but let me know if you think we should update to the latest preview instead

wli3:
Let's remove them. It is too hard to keep moving to new branches. And so far I don't see we missed anything

wli3:
@wtgodbe could you merge this change?

wtgodbe:
You could do what Asp.Net does to merge any 5.0-preview branch back to master:

https://github.com/dotnet/versions/blob/acc3af64db7225913b18a14dfc641f64acda85bf/Maestro/subscriptions.json#L1018-L1036

sfoslund:
Merging all 5.0-preview* branches to master sounds fine to me. Any objections, @wli3?

wli3:
let's see. If we have a last merge for them, it should be fine. If not, there will be a wave of PRs, and we have to resolve them (instead of close). And merge them. Then, they won't show up.

sfoslund:
We'll need to do those merges eventually anyway, so might as well get the bot to make them and the vendors to help out, right?

PR #680, opened 2021-02-19
Change auto merge from 5.0.2xx to 5.0.3xx

wli3:
Once 5.0.2xx reached servicing. It will get the latest runtime.
And that is not what we want in master. While 5.0.3xx is the branch with active
development


wli3:
@wtgodbe could you merge this?

@sfoslund FYI

PR #744, opened 2021-03-31
Fix dnceng services auto PR from production to main

riarenas:
Arcade-services' branch was renamed a couple of days ago.


riarenas:
Oh, this actually affects all of our repos, not just arcade-services. That's nice. All of the repos are now using main.

mmitche (approved):

PR #779, opened 2021-09-16
Update mono/linker -> dotnet/linker

sbomer:
I think the mirror hasn't been picking up changes since the repo was renamed. If approved I will also rename the internal repo to match.

@agocke PTAL, I can't seem to add reviewers.


agocke:
LGTM -- I didn't realize we also needed to change this repo.

sbomer:
@mmitche PTAL and merge if it looks good (I don't have permissions).

PR #782, opened 2021-09-29
Seed the microsoft/go-docker imageinfo files

dagood:
@mthalman Can you double-check my logic on these file names? (And is the amount of emptiness in the JSON correct?)

```
build-info/docker/image-info.microsoft-go-docker-main.json
build-info/docker/image-info.microsoft-go-docker-main-nightly.json
```

* This part seems standard: `build-info/docker/image-info.`
* This is based on the repo name, microsoft/go-docker: `microsoft-go-docker`
* The `-main` gets weirder.
  * The main branch in the repo is actually `microsoft/main`, and the nightly branch is planned to be `microsoft/nightly`. (Our repo is based on another one, so this prefix helps keep track of branches with changes for our Microsoft infra.)
  * The logic in https://github.com/dotnet/docker-tools/blob/main/eng/common/templates/steps/set-image-info-path-var.yml seems to check specifically for `nightly`, so both of our branches will hit the `else` and our `publicSourceBranch` will be `main` for both.
* The `-nightly` part would be specified by setting the `imageInfoVariant` var in our nightly pipeline only.

### DIFF
--- a/build-info/docker/image-info.microsoft-go-docker-main-nightly.json
+++ b/build-info/docker/image-info.microsoft-go-docker-main-nightly.json
@@ -1,0 +1,4 @@
+{
+  "schemaVersion": "1.0",
+  "repos": []
+}
--- a/build-info/docker/image-info.microsoft-go-docker-main.json
+++ b/build-info/docker/image-info.microsoft-go-docker-main.json
@@ -1,0 +1,4 @@
+{
+  "schemaVersion": "1.0",
+  "repos": []
+}


mthalman (commented):
The content of the files looks good.

The setting of the `publicSourceBranch` variable will be changing due to https://github.com/dotnet/dotnet-docker/pull/3088/files#diff-b71c402aa9515375604f8016bbb97e38d2155feb187c52c09788e70d37f98031.  You'll be able to customize exactly what it should be based on your branch names.  For example, for .NET, we'll have our own implementation at https://github.com/dotnet/dotnet-docker/pull/3088/files#diff-4e6efe6e9281b0049bba9bf603831c8b05c84157881a8cb0477da9f177b55e2c.

I guess it's a matter of whether you want to wait for those changes to get deployed or go with your approach now and rename the file later.  @MichaelSimons - Are you still planning to review https://github.com/dotnet/dotnet-docker/pull/3088?

dagood:
Thanks for taking a look, forgot about that PR shaking this up. It can wait.

PR #808, opened 2022-03-02
Add auto merge for the servicing branches of 6.0.Nxx

marcpopMSFT:
Source build works in the 1xx branch which can put us out of sync in servicing. Adding automatic flow from 1xx and 2xx up a level each.


marcpopMSFT:
@MichaelSimons as FYI.

PR #810, opened 2022-03-09
Fix broken github-dnceng-branch-merge-pr-generator subscription for sdk/installer

MichaelSimons:
The subscriptions added with https://github.com/dotnet/versions/pull/808 are not appearing to run.  I think this is because the `vsoSourceBranch`  is not correctly specified.  My understanding is the `vsoSourceBranch` is the branch for where the action lives (e.g. arcade) not where the pr generator should run

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -783,7 +783,7 @@
       ],
       "action": "github-dnceng-branch-merge-pr-generator",
       "actionArguments": {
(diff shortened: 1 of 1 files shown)


riarenas (approved):
I **think** this is correct. @mmitche would be able to confirm for sure.

marcpopMSFT (approved):

PR #820, opened 2022-04-22
Skip .2xx from now on

marcpopMSFT:
6.0.2xx goes out of support in May. The final changes should already be in the public repo. At this time, we should stop flowing changes from .1xx through .2xx and skip straight to .3xx.

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -777,25 +777,7 @@
         }
       }
     },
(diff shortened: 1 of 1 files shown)


marcpopMSFT:
FYI @MichaelSimons

MichaelSimons (approved):
LGTM

mmitche (approved):

PR #824, opened 2022-05-19
Update subscriptions for dotnet/icu

directhex:
We're following a new branching structure to avoid confusion with upstream branches.


directhex:
@dotnet/dnceng

premun:
Is this the same case like https://github.com/dotnet/versions/pull/823?

premun:
Specifically I am asking about whether icu is also small enough so that we don't care that Helix jobs from the repo won't go into the servicing queues

directhex:
Yep, exactly the same

directhex:
We don't run Helix tests on dotnet/icu

premun (approved):

PR #838, opened 2022-08-16
Add auto-merges for aspnetcore 7.0 branches

wtgodbe:
Adds auto-merges from rel/7.0-rc* to rel/7.0, and rel/7.0 to main.

### DIFF
--- a/Maestro/subscriptions.json
+++ b/Maestro/subscriptions.json
@@ -666,6 +666,39 @@
         }
       }
     },
+    // Automate merging aspnetcore release/7.0-rc branches back to release/7.0
+    {
+      "triggerPaths": [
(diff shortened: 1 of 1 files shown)


halter73 (approved):

Maestro/subscriptions.json
```diff
@@ -666,6 +666,39 @@
         }
       }
     },
+    // Automate merging aspnetcore release/7.0-rc branches back to release/7.0
+    {
+      "triggerPaths": [
+        "https://github.com/dotnet/aspnetcore/blob/release/7.0-rc/**/*"
```
dougbu, line 672: Will this try to merge release/7.0 into release/7.0❔ If yes, does that matter❔

wtgodbe: I don't think so, it's a direct copy of what runtime does & what we did last year

Issue #167, opened 2017-06-07
Add a list of dotnet/versions maintainers


dagood:
This would make it clear who can merge and who to ping for reviews. (I am one.)

dougbu:
addressed in #889

PR #889, opened 2023-05-30
Add CODEOWNERS

dougbu:


.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# Users referenced in this file will automatically be requested as reviewers for PRs that modify the given paths.
+# See https://help.github.com/articles/about-code-owners/
+
+*                   @dotnet/dnceng
+/build-info/docker/ @MichaelSimons @mthalman # dnceng will _not_ be asked to review docker/ files.
```
dougbu, line 5: @lbussell do you want the `write` role in this repo❔ At the moment, you're not visibly part of any team w/ that role and I therefore can't include you here.

mthalman: I think that would be good. Logan is doing most of the daily work for Docker now so that work may require some changes to this docker dir. In that case, can this just be set to `@dotnet/dotnet-docker-reviewers` instead?

dougbu: It's possible but not something that **Just Works:tm:**
Before I opened the PR, I had a version of this branch that I pushed. Since it showed the usual squiggly lines under the team, I decided to use the individuals but then `@lbussel` got the squigglies.
We could also give `@dotnet/dotnet-docker-reviewers` an explicit `write` role pretty easily. I can do that and update this PR if it's your preference&hellip;

mthalman: > We could also give @dotnet/dotnet-docker-reviewers an explicit write role pretty easily

That seems like the right thing to do.

dougbu: did the thing

dougbu:
oh, nice. no squigglies 😀

dagood:
I think we can say this closes https://github.com/dotnet/versions/issues/167. 🙂

dougbu:
> I think we can say this closes #167. 🙂

I should have waited until the 6th birthday of that bug before I opened this. Just missed that opportunity&hellip;

garath (approved):

PR #897, opened 2023-06-30
Mirror dotnet/main-14.x and 16.x instead of unversioned main

directhex:
Our workflow basically means a fresh branch for every upstream LLVM we track, and merging into a singular main branch is an absolute nightmare - better off to acknowledge the workflow, and mirror the named branches


dougbu:
1. why aren't 14.x and 16.x branches named e.g., release/14.x❔
2. your update changes code mirroring i.e., keeping identically-named branches synchronized between GitHub and the AzDO mirror repo. it won't change automatic merges e.g., to internal/* branches in AzDO was that your intent❔

directhex:
> 1. why aren't 14.x and 16.x branches named e.g., release/14.xgrey_question

We're forked from an upstream which uses `release/14.x` numbering. We need a way to differentiate between LLVM `release/14.x` and our patched `release/14.x`

> 2. your update changes code mirroring i.e., keeping identically-named branches synchronized between GitHub and the AzDO mirror repo. it won't change automatic merges e.g., to internal/* branches in AzDO was that your intent?

The intent is mirroring GH dotnet/release/main-16.x to AzDO dotnet/release/main-16.x and GH GH dotnet/release/main-14.x to AzDO dotnet/release/main-14.x (at least until we determine that we are definitely not going to revert the 14->16 change, at which point we can drop the 14 mirroring), instead of the existing GH dotnet/release/main to AzDO dotnet/release/main.

dougbu (approved):

PR #942, opened 2024-02-10
Add mirroring for the vscode-csharp prerelease branch

dibarbet:
We're going to be using the prerelease branch for creating builds, so needs to be mirrored to the internal repo.


dibarbet:
Don't believe I have permissions to merge, so whoever reviews please also merge (if it's good ofc)
Thanks!

dougbu (approved):

PR #967, opened 2024-05-23
Excluding some files from triggering builds

missymessa:


azure-pipelines.yml
```diff
@@ -1,8 +1,16 @@
 trigger:
 - main
+  paths:
+    exclude:
+    - subscriptions.json
```
dougbu, line 5: Are you sure this works❔ Might need to be
```suggestion
    - /Maestro/subscriptions.json
```

missymessa: Yup, looks like I was off by a directory

PR #987, opened 2024-07-01
Seed the microsoft/go-images image-info file

dagood:
* Redo of https://github.com/dotnet/versions/pull/782
* For https://github.com/microsoft/go/issues/157

@gdams 

/cc @mthalman @lbussell

PR #994, opened 2024-07-08
go-images image-info: populate full initial file

dagood:
* See https://github.com/dotnet/versions/pull/987
* For https://github.com/microsoft/go/issues/157

It turns out that `"repos": []` isn't enough to seed the file, and we got this error in a build:

```
-- Calculating new image info content
Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.InvalidOperationException: Removal of out-of-date content resulted in there being no content remaining in the target image info file. Something is probably wrong with the logic.
...
```

(From [PublishImageInfoCommand.cs#L217-L221](https://github.com/dotnet/docker-tools/blob/78d1b28945c95cfaea6cd01b6333500939fab552/src/Microsoft.DotNet.ImageBuilder/src/Commands/PublishImageInfoCommand.cs#L217-L221))

This PR adds the full file generated by [the build that hit that exception](https://dev.azure.com/dnceng/internal/_build/results?buildId=2489359&view=results) as the seed, instead.

@gdams 

/cc @mthalman @lbussell 

(Please merge upon approval--I don't have merge permissions. Thanks!)

### DIFF
--- a/build-info/microsoft/go-images/image-info.microsoft-go-images-main.json
+++ b/build-info/microsoft/go-images/image-info.microsoft-go-images-main.json
@@ -1,4 +1,1566 @@
 {
   "schemaVersion": "1.0",
-  "repos": []
+  "repos": [
+    {
+      "repo": "oss/go/microsoft/golang",
+      "images": [
+        {
+          "productVersion": "1.21",
+          "manifest": {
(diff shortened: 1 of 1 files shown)


mthalman (approved):
This is fine. But I think all that was needed was to stub out the repo, not all of the images. But this will work.

